PR #1, opened 2020-07-22
HTTP/3 Design Proposal

seanmonstar:
This includes a proposal for the `h3` crate, with the goal of providing a generic HTTP/3 implementation that can be used standalone, or integrated into hyper.

[Rendered](https://github.com/hyperium/h3/blob/rfc-init/design/PROPOSAL.md)

### DIFF
--- a/design/PROPOSAL.md
+++ b/design/PROPOSAL.md
@@ -16,7 +16,7 @@ What would be the goals of this effort, and what wouldn’t be.
     - Implementing QUIC also requires picking a TLS implementation. Various adopters may have different requirements for which TLS library they use.
 - Simple, optional integration in the `hyper` crate.
 - Performance and correctness are paramount.
-    - We should have a good way to frequently measure peformance.
+    - We should have a good way to frequently measure performance.
 - Collaborative with the community.
     - All current HTTP/3 Rust implementations have hard dependencies on their QUIC implementations. By making `h3` generic over any QUIC-like interface, we allow people to plug in to existing stacks. Meanwhile, `h3` providing the HTTP/3-level details can focus effort and attention.
     - The hyper brand has recognition as an open source HTTP project.
@@ -345,4 +345,3 @@ The performance of HTTP/3 is also critical. We need to setup both unit benchmark
     - Does this “feel” better to the Quinn developers?
     - However, by starting fresh, we can ask the Quinn developers for any “lessons learned” or “how would you start over”.
     - After analyzing quinn-h3, many design patterns would need to be changed to work with the above requirements, it may be easier to just rewrite.
-
--- a/design/PROPOSAL.md
+++ b/design/PROPOSAL.md
@@ -73,7 +73,6 @@ HTTP/3 is a new networking protocol that defines HTTP semantics specifically ove
 
 ### Non-goals
 
-- **Prioritization**: Stream priorities were defined in HTTP/2, and are especially useful for browsers to improve page load speeds. However, they are not specifically defined in QUIC or HTTP/3.
 - **Synchronization**: As much as possible, synchronization should not be an internal feature.
     - We’ve learned from building the `h2` crate, which includes a `Mutex` around the stream store and buffer, and many different stream handles. This adds contention, and makes it increasingly difficult to realize when a stream is no longer reachable.
 - **Runtime integration**: There shouldn’t be a dependency on any runtime. By not being opinionated about the runtime, a core library like an HTTP/3 state machine can see more adoption.
@@ -83,6 +82,9 @@ HTTP/3 is a new networking protocol that defines HTTP semantics specifically ove
 - **QPACK dynamic table**: the dynamic table is a performance optimization, but has the potential to block streams waiting on dynamic table updates. To reduce initial complexity, and to not require figuring out the correct heuristic, we can delay integrating the dynamic table till later. HTTP/3 even defaults to expecting zero space in the dynamic table unless the peer opts-in.
 - **Server Push**
 - **HTTP CONNECT** to tunnel over a single QUIC stream
+- **Prioritization**: Stream priorities were defined in HTTP/2, and are especially useful for browsers to improve page load speeds. However, they are not specifically defined in QUIC or HTTP/3. There is a [different proposal][http-prio] that we should consider once stabilized.
+
+[http-prio]: https://tools.ietf.org/html/draft-ietf-httpbis-priority-01
 
 
 ## 4. Public API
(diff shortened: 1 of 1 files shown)


LPardue:
Thanks for writing this up!

One note on prioritization, the QUIC and HTTP/3 specifications do not explicitly include priority signals but do mention the benefits of being able to locally prioritize streams when multiplexing. One of the most important factors is the default strategy when there is stream concurrency, for example an implementation might choose to send response data in FIFO, or multiplex all response data in round-robin fashion. @rmarx et al did some nice analysis work in this space and wrote it up at https://h3.edm.uhasselt.be/files/HTTP3_Prioritization_extended_3jul2019.pdf

HTTP/3 started with a priority scheme similar to HTTP/2 (weighted dependencies) but removed it due to some challenges. We are developing an alternative priority scheme, Extensible Priorities, in the IETF HTTP Working Group that is simpler, this is also back portable to HTTP/2. See https://tools.ietf.org/html/draft-ietf-httpbis-priority-01

While I can appreciate wanting to focus on an MVP, and punting on prioritization until it settles down, I'm not quite sure what the difference between `Non-goals` and `Eventually` sections are. I think there is some benefit to thinking about how a H3 layer would drive stream scheduling strategies in the underlying QUIC transport independent of what priority signals or scheme is used.

design/PROPOSAL.md
```diff
@@ -0,0 +1,348 @@
...
+## 1. Goals
+
+What would be the goals of this effort, and what wouldn’t be.
+
+- Provide a standalone HTTP/3 crate.
+    - This is similar to what we did with `h2`. This allows folks that want more control over HTTP/3 specifically to be able to do so. Or people may wish to reduce their dependencies to the bare minimum, and thus don’t want to include crates supporting HTTP/1 and 2 (like `hyper` does).
+    - We have the `h3` crate name already reserved, as one option.
+    - It shouldn’t decide on synchronization or runtime needs.
+- Allow users to provide their own QUIC implementation.
+    - There are several QUIC implmentations already, and new ones could appear in kernels instead of user-space.
```
jxs, line 15:
```suggestion
    - There are several QUIC implementations already, and new ones could appear in kernels instead of user-space. 
```

design/PROPOSAL.md
```diff
@@ -0,0 +1,348 @@
...
+- Provide a standalone HTTP/3 crate.
+    - This is similar to what we did with `h2`. This allows folks that want more control over HTTP/3 specifically to be able to do so. Or people may wish to reduce their dependencies to the bare minimum, and thus don’t want to include crates supporting HTTP/1 and 2 (like `hyper` does).
+    - We have the `h3` crate name already reserved, as one option.
+    - It shouldn’t decide on synchronization or runtime needs.
+- Allow users to provide their own QUIC implementation.
+    - There are several QUIC implmentations already, and new ones could appear in kernels instead of user-space. 
+    - Implementing QUIC also requires picking a TLS implementation. Various adopters may have different requirements for which TLS library they use.
+- Simple, optional integration in the `hyper` crate.
+- Performance and correctness are paramount.
+    - We should have a good way to frequently measure peformance.
```
jxs, line 19:
```suggestion
    - We should have a good way to frequently measure performance.
```

LPardue:
Something else to consider is logging. Many QUIC and HTTP/3 implementations have adopted qlog for structured logging, this can be fed into tools like qvis https://qvis.edm.uhasselt.be/#/files

Quiche has a [qlog sub-crate](https://crates.io/crates/qlog) that is used by quiche and neqo, it is mostly just the qlog datamodel combined with serde. I'm happy to take on board suggestions for that independent of any choices regarding the main quiche library.

seanmonstar:
> I'm not quite sure what the difference between `Non-goals` and `Eventually` sections are.

That's fair, the "Eventually" section grew after some previous reviews. Originally "non-goals" meant things to not figure out at first, but then I rearranged it a bit better. I agree the "prioritization" should be moved to "eventually, where as "non-goals" would continue to state things that are completely out of scope for the library, such as runtime integration.

davidbarsky:
> Quiche has a qlog sub-crate that is used by quiche and neqo, it is mostly just the qlog datamodel combined with serde. I'm happy to take on board suggestions for that independent of any choices regarding the main quiche library.

This is my bat signal! Given Hyper and H2's current usage of [`tracing`](https://github.com/tokio-rs/tracing), I suspect that `tracing` will be used for logging-like purposes. However, at a glance, the qlog data model appears to be similar to that of `tracing`'s, which means that it _shouldn't_ be too much work [knocks on wood] to create a `tracing-qlog` that integrates with [qlog](https://crates.io/crates/qlog).

(Why I might push for this approach is out of scope for this proposal.)

stammw:
I think there is a missing method for the `h3` impl to reject non-idempotent HTTP methods on zero-rtt requests. `quic::Connection::is_0rtt() -> bool` should be enough.

Defining custom streams for `quic` seem the best option to me. `AsyncRead` and `AsyncWrite` introduce a copy and force `std::Error` in the way so they should be avoided. One neat thing with them is the `tokio_util::codec::FramedRead` availability. I think having a decoder operating over `impl Buf for &[Bytes]` would be optimal.

I'd be OK for a fork, or we could move `quinn-h3` to `h3` directly. Then I'd be happy to continue working on it.

Previous state of `quinn-h3` was closer to your proposal, with external spawning and streaming bodies (through `AsyncRead` though). I don't think it would be take much efforts to reach a good fit on this API.

Ralith:
> [h3 works] without managing connections or spawning tasks

Note that this requires an application to respond to incoming connections with low latency even if it doesn't intend to begin doing work on them immediately, lest the peer experience an unpleasant, slow timeout. There is no middle ground like TCP where the connection can be accepted by the kernel long before the application sees it. This could be a bit of a footgun.

> Fork/rewrite quinn-h3

While my own focus is at the QUIC layer, I think I can say with some confidence that we'd like to avoid duplicating effort here, even if some refactoring is necessary.

LucioFranco:
> // Q: Should this be Pin<&mut Self>?

To me this should be a question of who we expect the consumer to be, if its internal (aka only hyper and h3 crates need to interact with it) then I think its fine to use pin here, that said, it would require `send_data` to also be `Pin<&mut Self>` or require that `impl SendStream` is always `Unpin`.

Overall from a high level, this looks great! I will have to continue reading up on h3/quic :)

seanmonstar:
Thanks for the feedback, all! I've updated some of the language, and confirmed with interested ones that things look good, so I'm merging and we'll see what happens!

PR #3, opened 2020-07-30
Add QUIC traits

seanmonstar:
This adds a `quic` module with a few traits, which can be used to
abstract over the QUIC transport.

### DIFF
--- a/src/quic.rs
+++ b/src/quic.rs
@@ -1,0 +1,106 @@
+//! QUIC Transport traits
+//!
+//! This module includes traits and types meant to allow being generic over any
+//! QUIC implementation.
+
+use std::task::{self, Poll};
+
+use bytes::Buf;
+
+// Unresolved questions:
+//
+// - Should the `poll_` methods be `Pin<&mut Self>`?
+
+/// Trait representing a QUIC connection.
+pub trait Connection<B: Buf> {
+    /// The stream type returned when opening a unidirectional send stream.
+    type SendStream: SendStream<B>;
+    /// The stream type returned when accepting a unidirectional receive stream.
+    type RecvStream: RecvStream;
+    /// The stream type returned when accepting or opening a bidirectional
+    /// stream.
+    type BidiStream: SendStream<B> + RecvStream;
+    /// The error type that can be returned when accepting or opening a stream.
+    type Error;
+
+    // Accepting streams
+
+    /// Poll the connection for any received bidirectional streams.
+    ///
+    /// Returning `None` implies the connection is closing or closed.
+    fn poll_accept_bidi_stream(
+        &mut self,
+        cx: &mut task::Context<'_>,
+    ) -> Poll<Result<Option<Self::BidiStream>, Self::Error>>;
+
+    /// Poll the connection for any received unidirectional streams.
+    ///
+    /// Returning `None` implies the connection is closing or closed.
+    fn poll_accept_recv_stream(
+        &mut self,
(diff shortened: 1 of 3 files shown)


src/quic.rs
```diff
@@ -0,0 +1,102 @@
...
+        &mut self,
+        cx: &mut task::Context<'_>,
+    ) -> Poll<Result<Option<Self::Buf>, Self::Error>>;
+
+    /// Send a `STOP_SENDING` QUIC code.
+    fn stop_sending(&mut self, error_code: u64);
+}
+
+/// Optional trait to allow "splitting" a bidirectional stream into two sides.
+pub trait BidiStream<B: Buf>: SendStream<B> + RecvStream {
```
WindSoilder, line 94: Hi, Would you please tell me that why `SendStream` use `Buf` as generic type,  and `RecvStream` use `buf` as associate type? 

It seems a little strange to me that they are not consistent

src/quic.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// The stream type returned when accepting or opening a bidirectional
+    /// stream.
+    type BidiStream: SendStream<B> + RecvStream;
+    /// The error type that can be returned when accepting or opening a stream.
+    type Error;
+
+    // Accepting streams
+
+    /// Poll the connection for any received bidirectional streams.
+    fn poll_accept_bidi_stream(
```
stammw, line 29: Maybe we could use an `Option<>` here to signify there won't be any new streams? Or should we assume there won't be any because `Self::Error::ConnectionClosed` or any other error ?

src/quic.rs
seanmonstar: Typically "input" types are generics on the trait, and "output" types are associated types. Since the user can "input" the buf type when sending, it's a generic.

src/quic.rs
seanmonstar: Er, yes, you're right! I even had that in the design proposal, but forgot it here.

src/quic.rs
```diff
@@ -0,0 +1,106 @@
...
+    type Error; // bounds?
+
+    /// Polls if the stream can send more data.
+    fn poll_ready(&mut self, cx: &mut task::Context<'_>) -> Poll<Result<(), Self::Error>>;
+
+    /// Send more data on the stream.
+    fn send_data(&mut self, data: B) -> Result<(), Self::Error>;
+
+    /// Poll to finish the sending side of the stream.
+    fn poll_finish(&mut self) -> Poll<Result<(), Self::Error>>;
```
stammw, line 71: `poll_finish()` needs a context object to be waken by the driver.

src/quic.rs
```diff
@@ -0,0 +1,106 @@
...
+    type Buf: Buf;
+    /// The error type that can occur when receiving data.
+    type Error; // bounds?
+
+    /// Poll the stream for more data.
+    ///
+    /// When the receive side will no longer receive more data (such as because
+    /// the peer closed their sending side), this should return `None`.
+    fn poll_data(
+        &mut self,
```
stammw, line 89: Should we make this `Pin<>` right now?

seanmonstar: The tricky part is, if this requires `Pin`, then how about the methods that wouldn't normally, such as `stop_sending`? That method itself likely doesn't need to be pinned, but once you've pinned to call `poll_data`, you can't easily unpin (generically) to fulfill `&mut self`... I don't know what the right answer is.

PR #4, opened 2020-08-02
Implement QUIC traits with Quinn

stammw:
Here is  a completely untested QUIC traits implementation with `quinn`.

It's based on #3 and contains some change suggestions to it. The draft status should disappear along with them.


seanmonstar:
I've merged #3. This looks really cool! I think we could have the integrations be in a separate crate, especially since this is creating new structs to implement the traits on anyways. We could create an `h3-quinn` directory with these types. I'm happy to do that busywork if you want.

stammw:
Yeah that makes more sense. Thanks!

stammw:
So I've got this sorted out with workspaces. I hope this what you meant ?

I also got rid of `Pin<&mut Self>` params.

h3-quinn/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+            return Poll::Ready(Ok(Some(chunk)));
+        };
+
+        ret
+    }
+
+    fn stop_sending(&mut self, error_code: u64) {
+        let _ = self
+            .stream
+            .stop(VarInt::from_u64(error_code).expect("invalid error_code"));
```
stammw, line 227: Do you think it's fine to panic if the error code is invalid here ? Or could we transfer the bound check to the user, accepting something like `VarInt` ?

seanmonstar: Yea, this was something I couldn't decide. I think there's these options (I could be missing some):

- Add some `h3::quic::ApplicationErrorCode` type that enforces the invariants. The downside is you'd probably need to convert *that* into a `u64` and then convert that into the QUIC-library-specific type, which means just check the value twice...
- Pass a `u64` and if it is bad, panic.
- Pass a `u64` and if it is bad, ignore it.
- Pass a `u64`, and if it is bad, return an error. That'd need us to change `stop_sending` to return `Result<(), Self::Error>`.

seanmonstar:
Awesome, yes this is what I had in mind! So it looks like we can't require Rust 1.39 as the minimum version (I just picked that since that's currently in hyper). Do you know what version Quinn does require?

h3-quinn/src/lib.rs
stammw: After all I guess setting it to `VarInt::MAX` if beyond the (sole) upper bound is a good trade-off. This is what I've just done. Normally, users should pass an `HttpError` enum value, and an unknown value does not represent a critical condition for the peer. 

This should be the broadest behavior, to fit most QUIC impls. 

Note that `VarInt` is not an only-QUIC thing. It is used for most integer values in h3 also: frame headers, stream types. And, it is a very simple type that as a upper bound check and is encoded with a "size hint" prefix on the most significant bits. It is not really an implementation-specific behavior.

`quinn-h3` uses `quinn::VarInt` and I think `h3` could as well. I'll open an issue for that matter.

stammw:
I don't know if there is a particular rust version target for `quinn`. `h3` being a major addition for `hyper`, I guess It wouldn't take much to bump the supported rust version along with its integration. This is an ultra-easy transition for the users IMHO.

seanmonstar:
Yea I'm not super worried about needing a newer version, I just like to have it recorded in CI, so we know whenever it needs to change. We can update the value in CI in this PR to make it happy with Quinn.

stammw:
Okay, so 1.40 is enough.
I had to tweak the doc checks in the CI a bit.

h3-quinn/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+    generic::{
+        Connection, IncomingBiStreams, IncomingUniStreams, OpenBi, OpenUni, RecvStream, SendStream,
+    },
+    ConnectionError, ReadError, VarInt, WriteError,
+};
+use quinn_proto::crypto::Session;
+
+use h3::quic;
+
+struct QuinnConnection<S: Session> {
```
jxs, line 22: shouldn't it be public, so that it can then be passed to `h3::server::Connection::handshake`?

stammw: Yes that should! Plus a `new(quinn::Connection)` function should be handy here. Thanks!

jxs: also NIT: since this is already under the quinn-h3 name could be just Connection and then let the importer rename if conflicting

stammw: Do you mean `QuinnConnection` should be named `Connection` ?

jxs: yeah, and then if/when naming conflicts occur, you can import with `use quinn_h3::Connection as QuinnConnection`

stammw: That's actually a lot better, I thought I would need those prefixes, but no.

stammw:
So we're fine to merge? Should I squash the commits as the green github button suggests ?

seanmonstar:
Squash or rebase, if the commits make sense separated. If the commits are just addressing review feedback, then they should probably be squashed.

PR #13, opened 2020-08-08
add hyper's BufList implementation

jxs:


stammw:
Nice! I'm not sure it has its place in `h3::proto` though. It will be used in `h3::frame` only.

jxs:
yeah i was not sure about it being in `h3::proto`, moved it to `h3::frame`

h3/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #[deny(missing_docs)]
 pub mod quic;
 
+mod frame;
```
stammw, line 4: Maybe we should silence all the "unused" warnings with temporary `#[allow(dead_code)]`. Then removed them once they are effectively used through public API.

h3/src/frame/mod.rs
```diff
@@ -0,0 +1 @@
+pub(crate) mod buf;
```
stammw, line 1: Let's keep this only accessible in the `frame` module for now.

Issue #9, opened 2020-08-06
Frame stream (reader)


stammw:
We want to poll a stream of [HTTP/3 frames] for the receiving ends of request streams, control stream and push streams. 

This is something I've already [implemented in `quinn-h3`]. This is based on `tokio_util::codec::FramedRead`, but will likely be easy to import and adapt to whatever solution we choose in #7 and #8. 

There is one point I'm not quite happy with though: the `DataFrame` partial handling. Here is how it works for now, imagine a response with a `2048` bytes body:

We recieve: `HeadersFrame(_) | DataFrame(payload { size: 2048 })`

```rust
let mut recv = FrameStream::new(recv_stream);

assert_matches!(recv.poll_next(cx).await, Some(Ok(HttpFrame::Headers(_))));

// Read the body
assert_matches!(
  recv.poll_next(cx).await,
  Some(Ok(HttpFrame::Data(DataFrame(b)))) if b.len() == 1024 // partial payload at the start (b: Bytes)
);
assert_matches!(
  recv.poll_next(cx).await,
  Some(Ok(HttpFrame::Data(DataFrame(b)))) if b.len() == 1024 // partial payload at the end
);

assert_matches!(recv.poll_next(cx).await, None); // No trailers
```

So as soon as there is some data available in the transport's buffer, it is yielded as a `Bytes` wrapped in a "fake" `DataFrame`. The problem here is that the user doesn't know what size the original `DataFrame` was. If we take a proxy use case, we will change the `DataFrame` sizes that we forward. If they are smaller than the original, that means we'll add one header for each of them.

What I have in mind is: 

```rust
let mut recv = FrameStream::new(recv_stream);

assert_matches!(recv.poll_next(cx).await, Some(Ok(HttpFrame::Headers(_))));

// Read the body
// iterate on "real" DataFrames
while let Some(Ok(HttpFrame::Data(ref mut payload))) = recv.poll_next(cx).await {
  // payload is a stream of bytes and the len of the original DataFrame
  assert_eq!(data.len(), 2048);
  assert_matches!(data.poll_data().await, Some(b) if b.len() == 1024); // b: Bytes
  assert_eq!(data.remaining(), 1024);
  assert_matches!(data.poll_data().await, Some(b) if b.len() == 1024); // b: Bytes
  assert_matches!(data.poll_data().await, None);
}

assert_matches(recv.poll_next(cx).await, None); // No trailers
```

[...]

PR #14, opened 2020-08-08
Implement FrameStream on top of BufList

### LINKED ISSUE #8: Framing codecs
This is the entry-point of the implementation effort: encoding / decoding frames as described in the [H3 draft].

A good thing is that we can almost directly import [quinn_h3::proto::frame]. But first, we need  to figure out a solution for #6, and possibly #7.

Also, it is based on [quinn's codec-like helpers] for `VarInt`s that make the job more ergonomic IMO. But that will be easy to import only the used parts.

[H3 draft]: https://quicwg.org/base-drafts/draft-ietf-quic-http.html#name-http-framing-layer
[quinn_h3::proto::frame]: https://github.com/djc/quinn/blob/master/quinn-h3/src/proto/frame.rs
[quinn's codec-like helpers]: https://github.com/djc/quinn/blob/master/quinn-proto/src/coding.rs

stammw:
closes #8 and addresses #9 partially.

### DIFF
--- a/h3-quinn/src/lib.rs
+++ b/h3-quinn/src/lib.rs
@@ -3,6 +3,8 @@
 //! This module implements QUIC traits with Quinn.
 use std::{
     collections::BTreeMap,
+    error::Error,
+    fmt::Display,
     task::{self, Poll},
 };
 
@@ -11,7 +13,7 @@ use futures::{ready, FutureExt, StreamExt};
 use bytes::{Buf, Bytes};
 use quinn::{
     generic::{IncomingBiStreams, IncomingUniStreams, NewConnection, OpenBi, OpenUni},
-    ConnectionError, ReadError, VarInt, WriteError,
+    ConnectionError, VarInt, WriteError,
 };
 use quinn_proto::crypto::Session;
 
@@ -235,6 +237,30 @@ impl<S: Session> quic::RecvStream for RecvStream<S> {
         let _ = self
             .stream
             .stop(VarInt::from_u64(error_code).expect("invalid error_code"));
+    }
+}
+
+#[derive(Debug)]
+pub struct ReadError {
+    cause: quinn::ReadError,
+}
+
+impl Display for ReadError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.cause.fmt(f)
+    }
+}
+
+impl Error for ReadError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        // TODO: implement std::error::Error for quinn::ReadError
+        None
+    }
+}
+
+impl From<quinn::ReadError> for ReadError {
+    fn from(e: quinn::ReadError) -> Self {
+        Self { cause: e }
     }
 }
 
--- a/h3/src/frame/buf.rs
+++ b/h3/src/frame/buf.rs
@@ -24,6 +24,24 @@ impl<T: Buf> BufList<T> {
     pub(crate) fn bufs_cnt(&self) -> usize {
         self.bufs.len()
     }
+
+    pub fn cursor(&self) -> Cursor<T> {
+        Cursor {
+            buf: &self,
+            pos_total: 0,
+            index: 0,
+            pos_front: 0,
+        }
+    }
+}
+
+#[cfg(test)]
+impl<T: Buf> From<T> for BufList<T> {
+    fn from(b: T) -> Self {
+        let mut buf = Self::new();
+        buf.push(b);
+        buf
+    }
 }
 
 impl<T: Buf> Buf for BufList<T> {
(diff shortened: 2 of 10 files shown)


h3/src/quic.rs
```diff
@@ -79,7 +82,7 @@ pub trait RecvStream {
     /// The type of `Buf` for data received on this stream.
     type Buf: Buf;
     /// The error type that can occur when receiving data.
-    type Error; // bounds?
+    type Error: Into<io::Error>;
```
stammw, line 85: @seanmonstar What do you think of this bound ? Quinn returns io::Error, and I think this fits the use case. 

It makes calling `.into()` twice in `h3::frame` though. So it makes me feel there is a better solution?

seanmonstar: Yea I wasn't sure, thus the question in comments. In hyper, I've required `Into<Box<dyn std::error::Error + Send + Sync>>`, which allows me to wrap the error with some context in hyper, and still provide access to the original in `Error::source()`.

I'd probably lean towards that, but wasn't sure if there was something better.

stammw: So we would require `dyn std::error::Error` if we want to add some context at the `h3` layer ? `Send + Sync` is also needed right ?

h3/src/frame/buf.rs
```diff
@@ -70,3 +70,41 @@ impl<T: Buf> Buf for BufList<T> {
         vecs
     }
 }
+
+pub struct Cursor<'a, B> {
```
jxs, line 74: do we need an own `Cursor`? `Buf` is impl'ed for `Cursor<T:Buf>`: https://docs.rs/bytes/0.5.6/bytes/trait.Buf.html#impl-Buf-for-Cursor%3CT%3E

stammw: Yeah, that's the problem. BufList content is not contiguous and therefore cannot implement `AsRef<[u8]>` as Cursor relies onto.

stammw:
Good thing I added some tests, a header or varint spread on multiple buffers results in an infinite loop!

h3/src/frame/buf.rs
```diff
@@ -70,3 +70,57 @@ impl<T: Buf> Buf for BufList<T> {
...
+
+impl<'a, B: Buf> Buf for Cursor<'a, B> {
+    #[inline]
+    fn remaining(&self) -> usize {
+        self.buf.remaining() - self.pos
+    }
+
+    #[inline]
+    fn bytes(&self) -> &[u8] {
+        &self.buf.bytes()[self.pos..]
```
seanmonstar, line 97: I think this won't work when used with a `BufList`, since calling `buf.bytes()` would only ever return the first buf's bytes.

h3/src/quic.rs
seanmonstar: Yes, If any, I'd add `type Error: Into<Box<dyn std::error::Error + Send + Sync>>`.

h3/src/proto/frame.rs
```diff
@@ -0,0 +1,543 @@
...
+    InvalidSettingValue(SettingId, u64),
+    SettingRepeated(SettingId),
+    SettingsExceeded,
+    Incomplete(usize),
+    IncompleteData,
+    Settings(String),
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Frame {
```
seanmonstar, line 25: Just a suggestion: perhaps we can remove the `Frame` suffix from the inner types (so, `Headers` instead of `HeadersFrame`, etc). Since we're already in the `frame` module, I think it will still be clear.

h3/src/quic.rs
stammw: Ok thanks. I'm not very knowledgeable when it comes to choose API bounds upfront for the moment.

h3/src/frame/buf.rs
stammw: I've written a special `Cursor` for `BufList`, now it doesn't go in an infinite loop anymore.

h3/src/proto/frame.rs
stammw: Done

h3/src/frame/mod.rs
```diff
@@ -1,3 +1,284 @@
...
+        let end = self.try_recv(cx)?;
+
+        let data = self.bufs.bytes();
+        match (data.remaining(), end) {
+            (0, true) => return Poll::Ready(Ok(None)),
+            (0, false) => return Poll::Pending,
+            (x, _) => *self.remaining_data.as_mut().unwrap() -= x as u64,
+        };
+
+        Poll::Ready(Ok(Some(Bytes::from(data))))
```
stammw, line 65: @seanmonstar How can I achieve this ? The borrow checker says it needs `'static`.

seanmonstar: To copy into a new `Bytes`, you'd need `Bytes::copy_from_slice(data)`. I think that's perfectly fine for now, let's get it working, and we can see about less copies in a separate PR.

h3/src/frame/mod.rs
```diff
@@ -1,3 +1,284 @@
...
+    }
+
+    pub fn reset(&mut self, error_code: ErrorCode) {
+        let _ = self.recv.stop_sending(error_code.0.into());
+    }
+
+    fn try_recv(&mut self, cx: &mut Context<'_>) -> Result<bool, Error> {
+        match self.recv.poll_data(cx) {
```
jxs, line 73: shouldn't `recv` be polled continuously until returning `Poll::Pending` as to [to register interest](https://doc.rust-lang.org/nightly/core/task/enum.Poll.html#variant.Pending)?
 otherwise it seems there's a state where `revc.poll_data()` returns `Some(d))` but `FrameDecoder` returns `None` due to the frame being `Incomplete` and `FrameStream` returns `Poll::Pending`  without setting to be woken when `recv` makes progress . 
See the [h2 example](https://github.com/hyperium/h2/blob/master/src/codec/framed_read.rs#L347) on how to keep polling until a `Frame` can be decoded

stammw: Looking closer into it, it looks like you're right! Thanks.

h3/src/quic.rs
stammw: Wouldn't `std::io::Error` have a large enough field of possibilities, instead of `std::error::Error` ?

seanmonstar: Typically `io::Error` is annoying, since the "cause" kind of *becomes* the `io::Error`, but only partly.

Doing as I said tends to be more flexible and clearer to users. But if it's easier, we can just merge with this and I can follow up with an error design PR.

stammw: I'll try to get this done in the current PR.

h3/src/frame/mod.rs
stammw: I think I got this sorted out. Could you check again ?

h3/src/frame/mod.rs
stammw: I quite unhappy with the solution I just came up with, but it should work.

h3/src/quic.rs
stammw: Done.

h3-quinn/src/lib.rs
```diff
@@ -238,6 +240,30 @@ impl<S: Session> quic::RecvStream for RecvStream<S> {
...
+
+impl Display for ReadError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.cause.fmt(f)
+    }
+}
+
+impl Error for ReadError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        // TODO: implement std::error::Error for quinn::ReadError
```
stammw, line 256: Do we need this ? Should I open a PR on quinn ?

seanmonstar: It's not required. It's only useful if users would want to be able to inspect for some specific variant or detail on the error.

seanmonstar (approved):
LGTM! Merge when ready :D

PR #16, opened 2020-08-28
add intial fuzzers to VarInt

jxs:
the seed corpus are imported from quiche's fuzz [`packet_recv_client` ](https://github.com/cloudflare/quiche/tree/master/fuzz)


h3/src/lib.rs
```diff
@@ -2,4 +2,4 @@
 pub mod quic;
 
 mod frame;
-mod proto;
+pub mod proto;
```
seanmonstar, line 5: We probably would want to keep this module private. Perhaps the fuzzing can be inside the crate... Or I suppose you use `#[path = "whatever"]`.

jxs: ah! sorry forgot to mention that, yeah `cargo-fuzz` expects a [fuzz repo](https://github.com/rust-fuzz/cargo-fuzz/blob/master/src/main.rs#L38), i have since updated it to use `#[path ]`.

stammw (approved):
Nice!

Issue #11, opened 2020-08-06
QPACK


stammw:
QPACK defines the way Headers are encoded on the wire. It has a [separate draft], but HTTP/3 simply cannot work without it. It is able to reuse already exchanged header names and values, by sending only reference to them in a dynamic table, which each peer maintain. An HTTP/3 connection as two unidirectional streams, one for each encoding directions: Encoder and decoder streams.

A valid implementation of it is ready to be imported from [`quinn-h3`]. It has some minor API design flaws though. @djc and I were about to [address them], but i'm not aware of any progress on this. Being the first bit I wrote, there are also some [identified improvement vectors]. Never the less, this part has a high test coverage but needs some fuzzing.

[separate draft]: https://quicwg.org/base-drafts/draft-ietf-quic-qpack.html
[`quinn-h3`]: https://github.com/djc/quinn/tree/master/quinn-h3/src/qpack
[address them]: https://github.com/djc/quinn/issues/228
[identified improvement vectors]: https://github.com/djc/quinn/issues?q=is%3Aissue+is%3Aopen+qpack

PR #17, opened 2020-09-24
Qpack

stammw:
Adresses #11 

This is an import of `quinn-h3`'s `qpack` module, with a few API adjustments: `Encoder` and `Decoder` structs instead of functions operating on `DynamicTable`s.

This doesn't contains refactorings mentionned in #11. Those can be defered after connection management implementation and benchmarking.


stammw:
Abandonning since included in #21

Issue #18, opened 2020-09-25
Headers / pseudo-headers management


stammw:
Import `quinn-h3`'s `headers` module. 

It makes the link between Raw decoded headers from `QPACK` and the `http::Resquet`, `http::Response` structs. Translating `:<pseudo-headers>` and enforcing validity for each of those header constraints.

PR #21, opened 2020-09-26
Headers

stammw:
Adresses #18


h3/src/qpack/stream.rs
```diff
@@ -0,0 +1,318 @@
...
+    InsertWithNameRef,
+    InsertWithoutNameRef,
+    Duplicate,
+    DynamicTableSizeUpdate,
+    Unknown,
+}
+
+impl EncoderInstruction {
+    pub fn decode(first: u8) -> Self {
+        if first & 0b1000_0000 != 0 {
```
seanmonstar, line 19: I'm guessing these constants are shared with the header block decoding bits? Would it make sense to consolidate them into a single set of constants?

h3/src/qpack/static_.rs
```diff
@@ -0,0 +1,215 @@
...
+    (b"server", b""),
+    (b"timing-allow-origin", b"*"),
+    (b"upgrade-insecure-requests", b"1"),
+    (b"user-agent", b""),
+    (b"x-forwarded-for", b""),
+    (b"x-frame-options", b"deny"),
+    (b"x-frame-options", b"sameorigin")
+];
+
+lazy_static! {
```
seanmonstar, line 154: I have a slight preference to *not* using a lazy static hashmap, if we can do similar to how we did it in h2: https://github.com/hyperium/h2/blob/2b19acf1323257e1d2a99ad9889e0619593561fd/src/hpack/table.rs#L677

Unless it's a huge performance difference, which I don't have any easy way to answer (so I just prefer less dependencies). What do you think?

seanmonstar (commented):
Awesome work! Just some thoughts inline.

h3/src/qpack/block.rs
```diff
@@ -0,0 +1,374 @@
...
+    IndexedWithPostBase,
+    LiteralWithNameRef,
+    LiteralWithPostBaseNameRef,
+    Literal,
+    Unknown,
+}
+
+impl HeaderBlockField {
+    pub fn decode(first: u8) -> Self {
+        if first & 0b1000_0000 != 0 {
```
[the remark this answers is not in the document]
stammw: Yeah, for sure! I've got these spec comments in 27d18ca. What do you think of these ? There are a lot of places where a similar work should be done.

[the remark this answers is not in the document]
seanmonstar: Yep, that's helpful, thanks!

h3/src/qpack/static_.rs
stammw: I've been looking into this, it's a pretty non-trivial change, and it could tie `qpack` with the `http` crate, but that's not really a problem. Regarding performance, it might even be faster, I guess.

h3/src/qpack/stream.rs
stammw: Actually not, this is another part of the spec.

h3/src/qpack/static_.rs
stammw: Okay, so it was far easier than I thought matching on byte slices...

h3/src/proto/coding.rs
```diff
@@ -0,0 +1,55 @@
+use bytes::{Buf, BufMut};
+
+use super::varint::VarInt;
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct UnexpectedEnd(pub usize);
+
+pub type Result<T> = ::std::result::Result<T, UnexpectedEnd>;
+
+pub trait Codec: Sized {
```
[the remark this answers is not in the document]
stammw: Done

stammw:
I added comments quoting the QPACK spec where I thought it was needed. 

I think this is ready for another review.

seanmonstar (approved):
Thanks, looks great to me! Squash and merge when ready!

Issue #19, opened 2020-09-25
Connection establishment


stammw:
Introduce `client` and `server` modules, with their respective `Connection` types. Then implement `handshake(quic::Connection) -> Connection` so a connection can be established.

For the moment `Connection` builders are not necessary, let's wait for settings and `ContolStream` implementations to introduce them.

PR #22, opened 2020-10-09
Connection

stammw:
Adresses #19.

There is a deviation from the proposal here: no handshake, but only construct functions. Sorry for not figuring this out before, but the handshake part is actually out of H3 scope. It is a QUIC matter. Plus, there is no exhange needed prior the first request. Once QUIC+TLS negociated the ALPN, we're ready to work with H3.

I'm not sure where to provide a `const ALPN = b"h3"`. It is something that will always be needed since there is no H3 without QUIC, and no QUIC without TLS.

### DIFF
--- a/h3-quinn/Cargo.toml
+++ b/h3-quinn/Cargo.toml
@@ -10,3 +10,7 @@ bytes = "0.5"
 quinn = "0.6"
 quinn-proto = "0.6"
 futures = "0.3"
+
+[dev-dependencies]
+tokio = { version = "0.2", features = ["rt-threaded", "macros"]}
+rcgen = "0.7.0"
--- a/h3-quinn/src/lib.rs
+++ b/h3-quinn/src/lib.rs
@@ -335,3 +335,71 @@ impl From<WriteError> for SendStreamError {
         Self::Write(e)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use h3::{client, server};
+    use quinn::{
+        Certificate, CertificateChain, ClientConfigBuilder, Endpoint, PrivateKey,
+        ServerConfigBuilder,
+    };
+    use rcgen;
+    use tokio;
+
+    #[tokio::test]
+    async fn connect() {
+        let (cert_chain, cert, key) = build_certs();
+        // Build client
+        let mut client_config = ClientConfigBuilder::default();
+        client_config.protocols(&[b"h3"]);
+        client_config.add_certificate_authority(cert).unwrap();
+
+        let mut client_endpoint_builder = Endpoint::builder();
+        client_endpoint_builder.default_client_config(client_config.build());
+        let (client_endpoint, _) = client_endpoint_builder
+            .bind(&"[::]:0".parse().unwrap())
+            .unwrap();
+
+        let client = async {
+            let conn = client_endpoint
+                .connect(&"[::1]:4433".parse().unwrap(), "localhost")
+                .unwrap()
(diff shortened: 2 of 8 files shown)


h3-quinn/Cargo.toml
```diff
@@ -10,3 +10,5 @@ bytes = "0.5"
 quinn = "0.6"
 quinn-proto = "0.6"
 futures = "0.3"
+tokio = { version = "0.2", features = ["rt-threaded", "macros"]}
```
seanmonstar, line 13: These could be `dev-dependencies`.

seanmonstar (commented):
>  no handshake, but only construct functions

My understanding of the HTTP/3 spec is that a few streams need to be setup first: the control streams and the qpack streams, and swapping of SETTINGS. That's what I meant by handshaking, at least. I sketched that out in an [early branch here](https://github.com/hyperium/h3/blob/spike/src/server.rs#L53-L81). I think 0rtt requests could skip it, but there might be value in knowing "all the necessary streams are ready". I could also be misunderstanding something in the spec, it's plenty big and I've not read it that many times.

stammw:
To me, recieving a `SettingsFrame` from the peer's control stream is not a hard requirement before exchanging requests. In [3.3. Connection Establishment](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#name-connection-establishment):

> While connection-level options pertaining to the core QUIC protocol are set in the initial crypto handshake, HTTP/3-specific settings are conveyed in the SETTINGS frame. After the QUIC connection is established, a SETTINGS frame (Section 7.2.4) MUST be sent by each endpoint as the initial frame of their respective HTTP control stream; see Section 6.2.1.

Only means `SETTINGS` must be *sent*.

As well as this in [6.2.1 Control Streams](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#control-streams), only says the peer *might* send it's `SETTINGS` first:

> Depending on whether 0-RTT is enabled on the QUIC connection, either client or server might be able to send stream data first after the cryptographic handshake completes.

I asked about this on the quicwg slack http channel:

stammw 
> Hello everyone!
> Must the control streams be opened before any request to be processed ? If so, must the peers wait for settings to be exchanged ? If no settings have been exchanged, how would the `MAX_HEADER_LIST_SIZE` be enforced ? By responding `431` if this unknown number is exceeded, that's all ?

evequefou

> Max header list size is an interesting one, because it's not actually normatively binding on anyone.  That reflects the fact that implementations have had such limits since HTTP/1.1 and simply not had a mechanism to disclose them.
> 
> So the setting is a way to early-fail -- the next hop down can not bother sending a request it knows will be rejected, rather than getting most(?) of the way through the headers and then getting the rejection.  But it's still perfectly valid to send that large request.
> 
> More generally, when you haven't seen SETTINGS yet, you're allowed to send but have to assume defaults.  That means no QPACK dynamic table and no negotiated extensions, primarily.
> 
> If you think the dynamic table would make your request substantially smaller, there's some advantage in waiting to see if you get a SETTINGS frame.  But there's also some advantage in not incurring any additional latency, so....  :man-shrugging: (edited) 

Yes we'd better open the outgoing control stream and send our settings on it. That will make the function taking care of this be `async`. But all the work here will only be *writting*. 

In this case I think `handshake()` is not a adequate name.

seanmonstar:
That's fair, thanks for the explanation. I'm fine with it just being `new` as an `async fn`.

seanmonstar (approved):

Issue #20, opened 2020-09-25
Client request sending


stammw:
Implement the `client::Connection` `send_request()` and `poll_ready()` methods.

`send_request()` will:
 - Convert `http::Request` to an `h3` specific `Header`
 - Encode this header with `QPACK` (in a stateless fashion for the moment, no dynamic table)
 - Open a `quic` bidirectional stream
 - Send the header on this stream
 - Return a `client::RequestStream<T::BidiStream, B> where T: quic::Connection<B>, B: Buf`

Let's address this `RequestStream` implementation in a later work, once server is able to receive and decode the header in order to emit a `server::RequestStream`.

seanmonstar:
Done in #23

PR #23, opened 2020-11-09
Request implementation

stammw:
Adresses #20.

I had to introduce stateless header compresison / decompression in `qpack::{en,de}code_stateless()` to prevent sharing state between `Connection` and `RequestStream`s for now. This code might also be useful later when we'll implement stateful dynamic QPACK.

I struggled to keep the `B: Buf` type argument of `RequestStream<C, B>` but couldn't help. I had to fix most of those to `Bytes`. I guess this is because the implementation manipulates buffers with `Bytes` and `BytesMut` in some places. Feels like I need some help figuring out what I'm missing here.

Maybe `send_data()` methods could accept `T: AsRef<&[u8]>` instread of `B` to make it more ergonomic ?

~~Some methods are exactly the same in `client` and `server` `RequestStream`, yet `send_response()` and `recv_request()` prevent to trivially de-duplicate those. I'll try to do this using another type parameter, like:~~ (DONE)

```rust
// Common impl block
impl<S, T> RequestStream<FrameStream<S>, Bytes, T>
where
    S: quic::RecvStream,
{
    pub async fn recv_data(&mut self) -> Result<Option<Bytes>, Error> {}
    pub async fn recv_trailers(&mut self) -> Result<Option<HeaderMap>, Error> {}
}

// Client's impl block
impl<S> RequestStream<FrameStream<S>, Bytes, Client>
where
    S: quic::RecvStream,
{
    pub async fn recv_response(&mut self) -> Result<Response<()>, Error> {}
}
```

I tried to reproduce `h2`'s tests organization. It's only helpful because `h3` and `h3-quinn` `quic` implementation are separated for now. But the same constraints as `h2` will likely start to show up at some point, so starting like this makes sense to me.

### DIFF
--- a/h3/src/proto/headers.rs
+++ b/h3/src/proto/headers.rs
@@ -21,10 +21,14 @@ pub struct Header {
 
 #[allow(clippy::len_without_is_empty)]
 impl Header {
-    pub fn request(method: Method, uri: Uri, fields: HeaderMap) -> Self {
-        Self {
-            pseudo: Pseudo::request(method, uri),
-            fields,
+    pub fn request(method: Method, uri: Uri, fields: HeaderMap) -> Result<Self, Error> {
+        match (uri.authority(), fields.get("host")) {
+            (None, None) => Err(Error::MissingAuthority),
+            (Some(a), Some(h)) if a.as_str() != h => Err(Error::ContradictedAuthority),
+            _ => Ok(Self {
+                pseudo: Pseudo::request(method, uri),
+                fields,
+            }),
         }
     }
 
--- a/h3/src/qpack/decoder.rs
+++ b/h3/src/qpack/decoder.rs
@@ -169,6 +169,41 @@ impl Decoder {
     }
 }
 
+// Decode a header bloc received on Request or Push stream. (draft: 4.5)
+pub fn decode_stateless<T: Buf>(buf: &mut T) -> Result<Vec<HeaderField>, Error> {
+    let (required_ref, _base) = HeaderPrefix::decode(buf)?.get(0, 0)?;
+
+    if required_ref > 0 {
+        return Err(Error::MissingRefs(required_ref));
+    }
+
+    let mut fields = Vec::new();
+    while buf.has_remaining() {
+        let field = match HeaderBlockField::decode(buf.bytes()[0]) {
+            HeaderBlockField::IndexedWithPostBase => return Err(Error::MissingRefs(0)),
+            HeaderBlockField::LiteralWithPostBaseNameRef => return Err(Error::MissingRefs(0)),
+            HeaderBlockField::Indexed => match Indexed::decode(buf)? {
+                Indexed::Static(index) => StaticTable::get(index)?.clone(),
+                Indexed::Dynamic(_) => return Err(Error::MissingRefs(0)),
+            },
+            HeaderBlockField::LiteralWithNameRef => match LiteralWithNameRef::decode(buf)? {
+                LiteralWithNameRef::Dynamic { .. } => return Err(Error::MissingRefs(0)),
+                LiteralWithNameRef::Static { index, value } => {
+                    StaticTable::get(index)?.with_value(value)
+                }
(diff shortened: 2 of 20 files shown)


h3/src/quic.rs
```diff
@@ -103,4 +106,7 @@ pub trait BidiStream<B: Buf>: SendStream<B> + RecvStream {
 
     /// Split this stream into two halves.
     fn split(self) -> (Self::SendStream, Self::RecvStream);
+
+    /// Get QUIC send stream id
+    fn id(&self) -> u64;
```
seanmonstar, line 111: This is already being added to `SendStream` above, so it shouldn't be needed here, I don't think.

h3/src/qpack/decoder.rs
```diff
@@ -169,6 +169,41 @@ impl Decoder {
...
+pub fn decode_stateless<T: Buf>(buf: &mut T) -> Result<Vec<HeaderField>, Error> {
+    let (required_ref, _base) = HeaderPrefix::decode(buf)?.get(0, 0)?;
+
+    if required_ref > 0 {
+        return Err(Error::MissingRefs(required_ref));
+    }
+
+    let mut fields = Vec::new();
+    while buf.has_remaining() {
+        let field = match HeaderBlockField::decode(buf.bytes()[0]) {
```
seanmonstar, line 182: I think I understand why, but just to double check: this doesn't use `buf.get_u8()` because that advances the buffer and below might return an error about not enough bytes yet, right?

h3/src/proto/headers.rs
```diff
@@ -21,10 +21,14 @@ pub struct Header {
 
 #[allow(clippy::len_without_is_empty)]
 impl Header {
-    pub fn request(method: Method, uri: Uri, fields: HeaderMap) -> Self {
-        Self {
-            pseudo: Pseudo::request(method, uri),
-            fields,
+    pub fn request(method: Method, uri: Uri, fields: HeaderMap) -> Result<Self, Error> {
+        match (uri.authority(), fields.get("host")) {
```
seanmonstar, line 25: Is the `host` header supposed to be a thing in HTTP/3? I think it's replaced by `:authority`, right?

seanmonstar (commented):
Phenomenal! Glad to see h3-tests too, looks like with this PR, things are working end-to-end! 🎉

h3/src/quic.rs
stammw: Yeah, plus this isn't needed as long as we don't have to track Dynamic table references in headers. I'll get rid of those for now.

h3/src/qpack/decoder.rs
stammw: We don't consume the first byte here because its first bits indicates the field type and the others are part of field data. So we leave advancing to the Header field decoder itself.

h3/src/proto/headers.rs
stammw: The `:authority` field is the preffered way of checking if the peer has the sufficient authority regarding its presented certificate. But `Host` can be used for that purpose:

In [ 4.1.1.1. Pseudo-Header Fields](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#name-pseudo-header-fields):

> If the ":scheme" pseudo-header field identifies a scheme that has a mandatory authority component (including "http" and "https"), the request MUST contain either an ":authority" pseudo-header field or a "Host" header field. If these fields are present, they MUST NOT be empty. If both fields are present, they MUST contain the same value. If the scheme does not have a mandatory authority component and none is provided in the request target, the request MUST NOT contain the ":authority" pseudo-header or "Host" header fields.

Do you want me to add this paragraph as a comment ?

seanmonstar: Oh cool, the spec is explicit about it. In the HTTP2 spec, it basically says host is dead unless coming from an HTTP/1 proxy. It's fine whichever you want to do here.

seanmonstar (approved):

PR #25, opened 2020-11-25
Create a client example

seanmonstar:
This starts creating usable examples so we can start seeing it interact with other implementations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,9 @@
 [workspace]
-members = ["h3", "h3-quinn", "tests/h3-tests"]
+members = [
+    "h3",
+    "h3-quinn",
+
+    # Internal
+    "examples",
+    "tests/h3-tests",
+]
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "examples"
+version = "0.0.0"
+publish = false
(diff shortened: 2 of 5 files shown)


seanmonstar:
@stammw I got the ALPN fixed (needed `h3-29` instead of just `h3` for several servers), but so far, they all timeout after sending the request. I'm probably missing something obvious?

seanmonstar:
Found the issue. After closing the send side, I also needed to add ignoring of unknown/reserved frames, and success!

stammw:
It actually works! Nice ! :)

seanmonstar:
I started adding printing of the received data too, and hit a panic `"There is still data to read, please call poll_data() ..."`...

seanmonstar:
Ok, so I fixed that by making sure `recv_data` wouldn't poll for a new frame if `FrameStream::remaining_data != 0`. Now I'm hitting a panic in h3-quinn, advancing the buffer too far after calling `read_unordered()`.

seanmonstar:
Ok, all set! `cargo run --example client https://blog.cloudflare.com` fetches the full page!

stammw:
All right! Sorry I haven't been that much involved into this but the my shiny new embedded linux video driver with fpga-cuda interface thing job is taking a lot of bandwidth.

PR #26, opened 2020-11-25
Treat reserved frame types as unknown frames

seanmonstar:
[Section 7.2.8](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-7.2.8) says that "reserved" frames are meant to be treated as unknown frames.


stammw (approved):

stammw:
Ok, no need to check if it is actually reserved types or not. Thanks !

PR #29, opened 2020-12-02
Fix h3-quinn receiving of ordered data

seanmonstar:
This fixes some bugs that existed due to the trickiness of unordered reads. With this change, I'm able to get the client example (#25) to download webpages.

### DIFF
--- a/h3-quinn/src/lib.rs
+++ b/h3-quinn/src/lib.rs
@@ -2,7 +2,6 @@
 //!
 //! This module implements QUIC traits with Quinn.
 use std::{
-    collections::BTreeMap,
     error::Error,
     fmt::Display,
     task::{self, Poll},
@@ -10,7 +9,7 @@ use std::{
 
 use futures::{ready, FutureExt, StreamExt};
 
-use bytes::{Buf, Bytes};
+use bytes::{Buf, Bytes, BytesMut};
 use h3::quic;
 pub use quinn;
 use quinn::{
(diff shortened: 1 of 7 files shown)


seanmonstar:
Ok, I'm hopeful this is the last piece to get the client streaming example working!

h3-quinn/src/lib.rs
```diff
@@ -203,38 +202,24 @@ impl<S: Session> quic::RecvStream for RecvStream<S> {
...
-            // If we get the chunk we're looking for, return it right away
-            Poll::Ready(Ok(Some((mut chunk, offset)))) if offset <= self.offset => {
-                chunk.advance((self.offset - offset) as usize); // XXX overflow
-                self.offset += chunk.len() as u64;
-                return Poll::Ready(Ok(Some(chunk)));
+        self.buf.resize(READ_BUF_SIZE, 0);
+
+        match self.stream.read(&mut self.buf).poll_unpin(cx) {
+            Poll::Ready(Ok(Some(n))) => {
+                eprintln!("stream read: {}", n);
```
stammw, line 209: `trace!()` should be used instead.

stammw (commented):
Nice work!

h3-quinn/src/lib.rs
seanmonstar: Yea woops, the first thing I did to debug was add these, before eventually adding a `tracing-subscriber` output to the unit tests.

stammw (approved):

PR #31, opened 2020-12-12
Fix unexpected ends when skipping a grease frame

seanmonstar:
I noticed trying to use the client example to connect to `https://quic.tech:8443` that it would fail with an `UnexpectedEnd` error. This happens because the stream data is small, and able to fit two GREASE frames and then a final DATA frame.

The previous `decode` would assume that since it was the final frame, but there was still data left, it had ended unexpectedly. The fix is to make `decode` loop since it may ignore unknown frames.


stammw (approved):

stammw:
Good catch ! Thanks!

PR #32, opened 2020-12-21
Change Error type to be opaque struct

seanmonstar:
This new design allows changing internal representation details without
it being a breaking API change. The design adds an `error::Code`, which
repesents an application error code scoped to HTTP/3 errors.

The design of the `Error` type has is fulfilling 2 roles:

- There was an underlying QUIC transport error.
- There was an HTTP/3 layer error, with a related code and possible
reason.


stammw (approved):
Yay! Integration improvement changes ! I didn't like how I've been handling errors so far. This impl is sweet !

Issue #36, opened 2020-12-30
Create a server example


seanmonstar:
We should create a simple "hello world" server example, that can be run via `cargo run --example server`.

PR #37, opened 2021-01-04
Server example

quentusrex:
Simplified example, that can either use self signed certificates, or if passed a cert chain and key, will use those instead.

### DIFF
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -1,0 +1,161 @@
+use h3_quinn::quinn::{Certificate, CertificateChain, PrivateKey};
+use std::path::PathBuf;
+use structopt::StructOpt;
+use tokio::fs::File;
+use tokio::io::AsyncReadExt;
+use tokio::stream::StreamExt;
+use tracing::{debug, error, info, trace, trace_span, warn};
+
+// Configs for two server modes
+// selfsigned mode will generate it's own local certificate
+// certs mode will require a path to 2-3 files(cert, key, ca)
+#[derive(StructOpt, Debug)]
+#[structopt(name = "server")]
+struct Opt {
+    #[structopt(subcommand)]
+    pub command: Command,
+}
+
+#[derive(StructOpt, Debug)]
+pub enum Command {
+    #[structopt(name = "selfsigned")]
+    SelfSigned(SelfSigned),
+
+    #[structopt(name = "certs")]
+    Certs(Certs),
+}
+
+#[derive(StructOpt, Debug)]
+pub struct SelfSigned {
+    #[structopt(long)]
+    pub debug: bool,
+
+    #[structopt(long, default_value = "4433")]
+    pub port: u16,
+}
+
+#[derive(StructOpt, Debug)]
+pub struct Certs {
+    #[structopt(long)]
+    pub cert: PathBuf,
+
+    #[structopt(long)]
+    pub key: PathBuf,
+
+    #[structopt(long, default_value = "4433")]
+    pub port: u16,
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -8,3 +8,6 @@ Cargo.lock
 
 # These are backup files generated by rustfmt
 **/*.rs.bk
+
+*~
+\#*
```
seanmonstar, line 13: Were these meant to be included?

examples/server.rs
```diff
@@ -0,0 +1,159 @@
...
+
+            let addr = format!("[::]:{:}", c.port);
+
+            server_endpoint_builder
+                .bind(&addr.parse().unwrap())
+                .unwrap()
+        }
+    };
+
+    warn!(
```
seanmonstar, line 107: Maybe just `info!`?

examples/server.rs
```diff
@@ -0,0 +1,159 @@
...
+        "Listening on port {:?}",
+        endpoint.local_addr().unwrap().port()
+    );
+
+    while let Some(new_conn) = incoming.next().await {
+        trace_span!("New connection being attempted");
```
seanmonstar, line 113: So that *this* connection doesn't block others, this should probably spawn a new task, `tokio::spawn`.

examples/server.rs
```diff
@@ -0,0 +1,159 @@
...
+        match new_conn.await {
+            Ok(conn) => {
+                debug!("New connection now established");
+
+                let (req, mut stream) =
+                    h3::server::Connection::new(h3_quinn::Connection::new(conn))
+                        .await
+                        .unwrap()
+                        .accept()
+                        .await
```
seanmonstar, line 123: So it looks like after accepting a connection, this will only accept 1 new request on it? After spawning a "connection" task, accepting a new request could probably be done in a loop and then likewise we'd spawn a new task to handle each request.

examples/server.rs
```diff
@@ -0,0 +1,159 @@
...
+#[derive(StructOpt, Debug)]
+#[structopt(name = "server")]
+struct Opt {
+    #[structopt(subcommand)]
+    pub command: Command,
+}
+
+#[derive(StructOpt, Debug)]
+pub enum Command {
+    #[structopt(name = "selfsigned")]
```
seanmonstar, line 21: Very cool!

.gitignore
quentusrex: emacs temp file patterns. I can strip them out, but they were useful to have while developing.

examples/server.rs
quentusrex: Ok, now it'll spawn the task for each new 'connecting' peer, then for each established peer, another task spawned when there is a request received.

examples/server.rs
quentusrex: I picked up this technique from the Tonari folks post here: https://blog.tonari.no/why-we-love-rust

It's made cli applications a TON easier.

.gitignore
quentusrex: @seanmonstar any preference here? I lean towards larger .gitignores, just to avoid accidental commits into a repo.

seanmonstar: I tend to think these files should be scoped to the specific project. It's always possible for any user to have a `git config --global core.excludesfile`, so you can prevent those files on your computer always.

quentusrex: Thanks. I wasn't aware of that feature.

quentusrex:
Linking PR to issue #36

seanmonstar (commented):
Awesome! Have you noticed testing with different clients (h3, curl, quiche, etc), any that send things that could cause the example server to hit some of the `unwrap`s? We'd probably want to not crash the server because of it, but if they barely happen in practice, I'm fine with the example merging without changing that. What do you think?

quentusrex:
I went through the different clients, and with the exception of the ALPN typo I had in the example server initially, I've not been able to crash this example server. I'm sure that a client fuzz would find some edge case that would trigger, the the ones I could replicate all have matches for the error cases.

seanmonstar (approved):
Woot!

Issue #33, opened 2020-12-22
Expose client and server RequestStreams


seanmonstar:
The current `connection::RequestStream` isn't publicly exported, and includes an extra type parameter to determine if you can send or receive a request. I'd suggest we should have 2 `RequestStream` types, one each in `client` and `server`, without the type parameter.

stammw:
I'll give a go at this!

seanmonstar:
Done in #40

PR #40, opened 2021-01-09
Expose client and server RequestStreams

stammw:
Addresses #33

### DIFF
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -1,11 +1,10 @@
 use bytes::{Bytes, BytesMut};
 use futures::future;
-use http::{request, Response};
+use http::{request, HeaderMap, Response};
 use std::convert::TryFrom;
 
-use crate::connection::RequestStream;
 use crate::{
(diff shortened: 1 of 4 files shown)


h3/src/client.rs
```diff
@@ -65,14 +66,20 @@ where
     }
 }
 
-impl<S, C> RequestStream<FrameStream<S>, Bytes, Connection<C>>
+pub struct RequestStream<S, C>
+where
+    C: quic::Connection<Bytes>,
```
seanmonstar, line 71: I believe this generic is to allow internal `connection::RequestStream` to know if it can `send_request` or `accept`. If that's the only thing, it'd be more convenient I think for users if we could just remove this generic, and not worry about enforcing that internally, only via the separate public `RequestStream` types.

stammw: Yeah, this wasn't obvious how to remove this at the time. Turns out it is useless :sweat_smile:

seanmonstar (approved):
Excellent!

Issue #39, opened 2021-01-07
Upgrade to Bytes v1.0


seanmonstar:
We should upgrade `h3` to depend on `bytes` v1.0.

PR #41, opened 2021-01-15
Update quinn and h3-29

zz85:
Currently running the client against popular endpoint such as google.com would result in a version mismatch failure. Using the updated version of quinn fixes it.

```
client https://google.com
DNS Lookup for https://google.com/: [2607:f8b0:400a:803::200e]:443
QUIC connected ...
Sending request ...
Receiving response ...
Response: HTTP/3.0 301 Moved Permanently
Headers: {
    "location": "https://www.google.com/",
    "content-type": "text/html; charset=UTF-8",
    "date": "Fri, 15 Jan 2021 06:07:46 GMT",
    "expires": "Sun, 14 Feb 2021 06:07:46 GMT",
    "cache-control": "public, max-age=2592000",
    "server": "gws",
    "content-length": "220",
    "x-xss-protection": "0",
    "x-frame-options": "SAMEORIGIN",
    "alt-svc": "h3-29=\":443\"; ma=2592000,h3-T051=\":443\"; ma=2592000,h3-Q050=\":443\"; ma=2592000,h3-Q046=\":443\"; ma=2592000,h3-Q043=\":443\"; ma=2592000,quic=\":443\"; ma=2592000; v=\"46,43\"",
}
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="https://www.google.com/">here</A>.
</BODY></HTML>
```

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -39,7 +39,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.42.0
+          - 1.45.0
 
     runs-on: ubuntu-latest
 
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -35,7 +35,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // quinn setup
     let mut client_config = h3_quinn::quinn::ClientConfigBuilder::default();
-    client_config.protocols(&[b"h3-27"]);
+    client_config.protocols(&[b"h3-29"]);
 
     let mut client_endpoint_builder = h3_quinn::quinn::Endpoint::builder();
     client_endpoint_builder.default_client_config(client_config.build());
--- a/h3-quinn/Cargo.toml
+++ b/h3-quinn/Cargo.toml
@@ -7,8 +7,8 @@ edition = "2018"
 [dependencies]
 h3 = { path = "../h3" }
 bytes = "0.5"
-quinn = "0.6"
-quinn-proto = "0.6"
+quinn = { git ="https://github.com/quinn-rs/quinn.git", branch="main" }
+quinn-proto = { git ="https://github.com/quinn-rs/quinn.git", branch="main" }
 futures = "0.3"
 
 [dev-dependencies]


seanmonstar:
Hmph, looks like something needs a newer `rustc` *grumble*. I guess we can just bump CI to the same minimum Rust as Tokio: 1.45?

seanmonstar:
I've filed https://github.com/quinn-rs/quinn/issues/979 to see if Quinn can support 1.45 too. While that may be important for whatever eventually gets included in hyper or reqwest, it might be that other QUIC libraries don't want to support older rustc either. So, I wonder if it's worth changing it so that CI only checks `h3` directly, and not integration crates, for minimum Rust version support. Thoughts?

stammw:
Do you mean checking only the `h3` crate for compilation success? I think that getting integration crates out of the CI would make `h3` a lot harder to validate ? We'd have to write a fake implementation (from what I understand). 

We could have other CI workflows for the integration crates, but that would lead to the same problems in the end? 

I've never been working on this kind of considerations...

seanmonstar:
Sorry, I realize I didn't write what I meant. I still would want CI to check `h3-quinn`, `examples`, and `tests`. I meant changing CI so that the MSRV (Rust 1.45 currently) only checks `h3`. The rest of the targets would still check all the rest.

I don't love what I'm proposing, just noting that the current situation doesn't allow us to use the latest Quinn and keep MSRV happy. Another fix would be convincing the Quinn maintainers to support 1.45.

seanmonstar:
Ok, looks like Quinn now supports Rust 1.45, but they've also upgraded to Tokio 1.0, so we need to as well.

stammw:
So now our problem is on hyper side ?

seanmonstar:
Yea, in order to support Quinn main, we need to update `h3-quinn` (and examples/tests) to use Tokio 1.0. And #39.

seanmonstar:
Alright, so as part of the update to Tokio 1.0 in #42, that included an update for Quinn. So probably this can just be an upgrade to a newer ALPN string?

stammw:
That's actually something to think about a bit, as no all quic impl will offer the same draft level. At least as long as quic + h3 are not released. It could even be a use case for future version iterations after that.

seanmonstar:
Thanks for this, we finally got updated to new Tokio and Quinn, which provided the full support we needed.

PR #47, opened 2021-04-12
Accept unidirectionnal streams and ControlStream impl

stammw:
This PR enables uni-directionnal streams support.

It brings the "connection driving" shared state problem onto the table. This is why the two last commits are marked as DRAFTs. 
I took inspiration from h2 to manage Request streams (bi-directionnal) and connection-control stream (uni-directionnal). Dispatch them into tasks without lock / synchronization. 

On the server side, the "accept requests" task can be assumed to be continuously polled, which makes avoiding locks easy. But on the client side, to open new requests streams and poll for incoming recv streams, we need to share a mutable reference on the `Connection`.

@seanmonstar I recall you'd like to avoid using any locks in this crate, should I let `hyper` manage the concurrent accesses to the client Connection ?

### DIFF
--- a/h3/src/quic.rs
+++ b/h3/src/quic.rs
@@ -13,47 +13,68 @@ use bytes::Buf;
 
 /// Trait representing a QUIC connection.
 pub trait Connection<B: Buf> {
-    /// The stream type returned when opening a unidirectional send stream.
-    type SendStream: SendStream<B>;
-    /// The stream type returned when accepting a unidirectional receive stream.
-    type RecvStream: RecvStream;
-    /// The stream type returned when accepting or opening a bidirectional
-    /// stream.
-    ///
-    /// This type doesn't *have* to implement `BidiStream`, but if the
-    /// implementation is able to provide a cheap split, it can opt-in.
+    /// The type produced by `poll_accept_bidi()`
     type BidiStream: SendStream<B> + RecvStream;
-    /// The error type that can be returned when accepting or opening a stream.
+    /// The type of the sending part of `BidiStream`
+    type SendStream: SendStream<B>;
+    /// The type produced by `poll_accept_recv()`
+    type RecvStream: RecvStream;
+    /// A producer of outgoing Unidirectional and Bidirectional streams.
+    type OpenStreams: OpenStreams<B>;
+    /// Error type yeilded by this trait methods
     type Error: Into<Box<dyn std::error::Error + Send + Sync>>;
 
-    // Accepting streams
-
-    /// Poll the connection for any received bidirectional streams.
+    /// Accept an incoming unidirecional stream
     ///
     /// Returning `None` implies the connection is closing or closed.
-    fn poll_accept_bidi_stream(
-        &mut self,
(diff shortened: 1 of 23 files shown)


seanmonstar:
> I recall you'd like to avoid using any locks in this crate, should I let hyper manage the concurrent accesses to the client Connection?

That's certainly one way! I realize having *zero* locks in Rust might get tricky, due to Rust's single-owner preference. I think by trying to come up with designs without them, we'll be better off, but if small areas need one, I think that's okay too. Ideally, we don't follow the mistake of the `h2` crate and have a huge chunk of state that needs to be locked for everything all the time. We've noticed there's a bunch of contention on that lock in `h2` (there's a desire to refactor to reduce contention).

h3/src/quic.rs
```diff
@@ -96,6 +96,17 @@ pub trait RecvStream {
         cx: &mut task::Context<'_>,
     ) -> Poll<Result<Option<Self::Buf>, Self::Error>>;
 
+    /// Poll the stream for a precise number of bytes
+    ///
+    /// The `buf` slice should be filled with as much received data as possible. When
+    /// the receive side will no longer receive more data (such as because
+    /// the peer closed their sending side), this should return `None`.
+    fn poll_read(
```
seanmonstar, line 104: Why the inclusion of this new method, instead of using `poll_data`?

seanmonstar (commented):
Looks great at a high level. I think I only have 1 main question, left inline.

h3/src/quic.rs
stammw: `poll_data()` reads everything available, but [here](https://github.com/hyperium/h3/pull/47/files/cc0e3a0e21447b4f2e561b52d78d43adb9eee503#diff-0845c3b9e922cf769bbc50f9ecaf7abc30f9b57bfaa73b1d4fd4b13c758e7e91R103) we want to consume only the first `VarInt` of the new stream.

stammw:
Sorry for letting this PR inactive for so long. If we want to avoid locks on the client, some API design change will be necessary: split `Connection` into an outgoing Request opener and a unidirectional stream listener (and reader, for the connection-wide streams, i.e. control, encoder, decoder) that would be polled from the driver task. Similar to what `quinn` does witth `IncomingBidi` and `IncomingUni`. I'll try to come up with an implementation of this soon.

stammw:
I'm finally happy with this PR.

It now contains changes that make it possible to share the minimal amount of state behind a `RwLock`. This state is only written when control stream receive some frames (Settings for now, Goaway and others in the future), which happens only a few specific times in the connection life-cycle. The rest of the time, it can be accessed in parallel by all the requests: they do not write to it, so there is no contention at their level.

I had to split the quic connection trait into Incoming and Outgoing sides to be able to drive the client connection in one task and send requests in another. This plays very well with Quinn, but will force the other QUIC implementation that didn't foresee this use-case to hold a lock in their h3::quic traits implementations.

stammw:
Superseded by #54

PR #56, opened 2021-11-02
conform to quinn api changes

FrankSpitulski:
quinn had some api changes, so I've updated h3 to work with the new api on the master branch. marked as WIP since there hasn't been an 0.8.0 release with the new api yet.


stammw:
Nice! Looks like the transport init code is a lot simpler now! I'll test this soon.

Did you have a look in tests ? I think helpers might need some changes too ?

FrankSpitulski:
I updated the examples, but it seemed to compile fine as is (which was my goal to make it work with my downstream project). 

Since you’ve got a major refactor coming in I think it can just wait to see which drops first, quinn 0.8.0 or the h3 refactor.

FrankSpitulski:
@stammw I rebased and updated the tests package. the "test connect" test is failing now with a generic error. "h3::Error { code: H3_STREAM_CREATION_ERROR, cause: ApplicationClosed(ApplicationClose { error_code: 0, reason: b"" }) }"

I tried working out the exact issue but I'm very unfamiliar with the manual async style polling code. On my real world application though it seems to work just fine with the update.

stammw:
All right, I'll have a look at this.

stammw:
Could you use `cargo fmt --all` to fix the style ?

h3/src/lib.rs
```diff
@@ -6,7 +6,7 @@ pub mod server;
 
 #[allow(dead_code)]
 mod connection;
-mod frame;
+pub mod frame;
```
stammw, line 9: Is this change needed in the current PR ?

stammw (changes requested):
Almost ready! Can you please make sure other server implementations still answer the client example ?

stammw:
And also: 

Now the CI doesn't pass because the new version of rustls uses combinators on booleans. Can you add a commit like this one https://github.com/g2p/rustls-native-certs/commit/0e1a0134c2b6ac26467fa1fa11e27d1b7aa7408d ?

examples/client.rs
```diff
@@ -50,34 +54,24 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     eprintln!("DNS Lookup for {:?}: {:?}", dest, addr);
 
     // quinn setup
-    let mut client_config = h3_quinn::quinn::ClientConfigBuilder::default();
-    client_config.protocols(&[ALPN]);
```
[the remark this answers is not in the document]
FrankSpitulski: yes, interop has now been fixed

h3/src/lib.rs
FrankSpitulski: probably not necessary to update to quin 0.8, but I needed it because I passed the connection as a method argument and had to type the fully qualified path out in my downstream project. since it wasn't public it wasn't letting me do that.

FrankSpitulski: to be clear, I have this line written out and used as arguments throughout my program. 

`type RequestStreamTyped = RequestStream<FrameStream<BidiStream<Bytes>>>;`

stammw: The API is not very ergonomic :thinking:

FrankSpitulski: since it’s not integrated with hyper or warp yet I wanted to pass the stream to my own hand rolled request matcher function. if I did it inline, not requiring the full definition, then I would have been 5 of 6 indents in. I never needed to interact with the frame itself, so maybe this meta type could just be exported in h3.

FrankSpitulski: If we’re talking about ergonomics though, the body being hard coded as () and then using a separate method for the real body is weird. It seems like you should be able to specify one or the other. It took me a few tries to figure out I was supposed to ignore the () body and use the other method.

stammw:
Did you check that server and client examples still work together ?

h3/src/lib.rs
stammw: Regarding the body, you are right. But if you want to stream it, using the `Request<>` associated type makes it kind of hard.
For now there is no docs because the API is subject to change. We need to find an API that is suitable for Hyper (and Actix), as well as ergonomic. Even if this crate is low-level.

I'll open an issue today to address the `FrameStream<>` ergonomics problem separately.

Thanks a lot for your feedback, it is very valuable. BTW, is your project available somewhere, so I can have a look ?

stammw:
The only missing thing is the client example against server example case. Then we can merge it.

It appears broken:

```
❯ cargo run --example=server selfsigned
```

```
❯ cargo run --example=client -- --insecure https://localhost:4433
Error: TransportError(Error { code: Code::crypto(78), frame: None, reason: "ALPN negotiation failed" })
```

h3/src/lib.rs
FrankSpitulski: unfortunately I can’t share the downstream project, but a relatively standard http3 server with a request router method and and auth filter in front of it would get you close in terms of the h3 interaction. Since you’re going to address the issue separately would you like me to remove the pub from here?

.github/workflows/CI.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           - stable
           - beta
           - nightly
-          - 1.46
+          - 1.50
```
stammw, line 42:
```suggestion
          - "1.50"
```

stammw (commented):
Looks like Github actions takes this as a float and removes the trailing `0`...

stammw (approved):

stammw:
Thanks!

Issue #57, opened 2021-11-24
Improve `RequestStream`s ergonomics


stammw:
As discussed in #56, it is not possible to name the full `RequestStream` type without `frame::FrameStream` being externally accessible. Namely, the return value of: 

```rust
   // h3/src/client.rs
    pub async fn send_request(
        &mut self,
        req: http::Request<()>,
    ) -> Result<RequestStream<FrameStream<C::BidiStream>>, Error> {
```

Cannot be stored in a struct or passed around in functions params without `FrameStream` being public.

I see multiple solutions to this:
- The best one in my opinion: extract a trait and return an `impl RequestStream`. 
  - Pros:
    - Makes the type opaque, so future proof in case of API changes.
  - Cons:  
    - That's not possible until async traits are a thing.
    - Users will have to `Box` it if they need to store this in a struct.
- Make `frame` public or reexport `FrameStream` in the crate's scope.
  - Pros:
    - It's easy to implement
  - Cons: 
    - The type is still quite long to write in downstream code
    - `FrameStream` is externally available but will never be used directly
- Create type aliases like `type RequestStream = RequestStreamImpl<FrameStream<C::BidiStream>>`
   - Pros:
     - It's easy to implement
   - Cons:
     - It would introduce another type if we want to split the streams into a reader and a writer: `type SendStream = RequestStreamImpl<C::SendStream<B>>`

stammw:
Fixed in #63

PR #63, opened 2021-11-27
Hide `FrameStream` from public API

stammw:
As dicussed in #56 and #57, wrap the `RequestStream` impls to remove the need of importing `FrameStream` to name the type. I used `Deref` and `DerefMut` to make all the methods transparently callable in downstream code. I wonder how recommendable is that technique?

I also updated the examples to show how to pass `RequestStream`s as argument, including the boundaries to use.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -115,10 +115,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
         while let Some(chunk) = stream.recv_data().await? {
             let mut out = tokio::io::stdout();
-            out.write_all(&chunk).await?;
-            out.flush().await?;
+            out.write_all(&chunk).await.expect("write_all");
+            out.flush().await.expect("flush");
         }
-        Ok::<(), Box<dyn std::error::Error>>(())
+        Ok::<_, Box<dyn std::error::Error>>(())
     };
 
     tokio::select! {
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -44,7 +44,7 @@ where
     pub async fn send_request(
         &mut self,
         req: http::Request<()>,
-    ) -> Result<RequestStream<FrameStream<T::BidiStream>>, Error> {
+    ) -> Result<RequestStream<T::BidiStream>, Error> {
         let peer_max_field_section_size = self
             .conn_state
             .read("send request lock state")
@@ -209,17 +209,23 @@ where
     }
 }
 
-pub struct RequestStream<S> {
-    inner: connection::RequestStream<S, Bytes>,
+pub struct RequestStream<S>
+where
+    S: quic::RecvStream,
+{
+    inner: connection::RequestStream<FrameStream<S>, Bytes>,
 }
 
-impl<S> ConnectionState for RequestStream<S> {
+impl<S> ConnectionState for RequestStream<S>
+where
+    S: quic::RecvStream,
+{
     fn shared_state(&self) -> &SharedStateRef {
         &self.inner.conn_state
     }
 }
 
-impl<S> RequestStream<FrameStream<S>>
+impl<S> RequestStream<S>
 where
     S: quic::RecvStream,
 {
@@ -281,7 +287,7 @@ where
 
 impl<S> RequestStream<S>
 where
-    S: quic::SendStream<Bytes>,
+    S: quic::RecvStream + quic::SendStream<Bytes>,
 {
     pub async fn send_data(&mut self, buf: Bytes) -> Result<(), Error> {
         self.inner.send_data(buf).await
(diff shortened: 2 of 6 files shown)


stammw:
@FrankSpitulski With this PR, the `FrameStream` doesn't have to be visible in downstream code anymore.

Still, a type alias needs to be defined if brevity at arguments passing or struct definition sites is desired.

FrankSpitulski:
LGTM

h3/src/quic.rs
```diff
@@ -29,7 +29,7 @@ impl<'a, E: Error + 'a> From<E> for Box<dyn Error + 'a> {
 /// Trait representing a QUIC connection.
 pub trait Connection<B: Buf> {
     /// The type produced by `poll_accept_bidi()`
-    type BidiStream: SendStream<B> + RecvStream;
+    type BidiStream: BidiStream<B>;
```
seanmonstar, line 32: The `BidiStream` trait is an opt-in, in case the QUIC library is able to support splitting them. The goal is that splitting shouldn't be _required_.

h3/src/client.rs
```diff
@@ -209,17 +210,40 @@ where
     }
 }
 
-pub struct RequestStream<S> {
+pub struct RequestStream<T>(RequestStreamImpl<FrameStream<T>>)
+where
+    T: quic::RecvStream;
+
+impl<T> Deref for RequestStream<T>
```
seanmonstar, line 217: It's a strongly held belief in the Rust community that `Deref` should _only_ be used for smart pointers, not for extending an inner type.

h3/src/client.rs
```diff
@@ -209,17 +210,40 @@ where
...
+impl<T> DerefMut for RequestStream<T>
+where
+    T: quic::RecvStream,
+{
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+pub struct RequestStreamImpl<S> {
```
seanmonstar, line 236: Instead of making a wrapper type, does the generic need to contain `FrameStream<T>` at all? Is there ever a use where a different generic is used internally? If not, you might be able to just have it be:

```rust
pub struct RequestStream<T> {
    inner: connection::RequestStream<FrameStream<T>, Bytes>,
}
```

h3/src/quic.rs
stammw: Thanks, I understood it the other way around :sweat_smile:

h3/src/client.rs
stammw: Thanks, removed.

h3/src/client.rs
stammw: Why didn't I even think of that !

stammw:
Thanks for the mentoring !

examples/client.rs
```diff
@@ -107,18 +109,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        eprintln!("Response: {:?} {}", resp.version(), resp.status());
-        eprintln!("Headers: {:#?}", resp.headers());
-
-        while let Some(chunk) = stream.recv_data().await? {
-            let mut out = tokio::io::stdout();
-            out.write_all(&chunk).await?;
-            out.flush().await?;
-        }
-        Ok::<(), Box<dyn std::error::Error>>(())
+        Ok::<_, Box<dyn std::error::Error>>(handle_request(stream).await)
```
seanmonstar, line 112: I think this makes it a `Result<Result<(), E>, E>`.

examples/client.rs
```diff
@@ -133,6 +124,23 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+async fn handle_request<T: quic::BidiStream<Bytes>>(
```
seanmonstar, line 127: This is just my thoughts as I read this, feel free to disagree with me. But I feel the example was easier to read without the extra function. At least as an initial first glance, "how do I do anything with the client". I'd consider showing something like this `handle_request` function in the documentation for the `client` module or something.

🤷

examples/client.rs
stammw: If rustc says this compiles it means everything is all right isn't it :smile:. I remove the function.
Thanks.

examples/client.rs
stammw: You are right, it adds complexity to show something that belongs to the docs.

seanmonstar:
Otherwise looks good! I'm happy to approve and merge when ready (in case you're still working, or just didn't push the branch).

stammw:
Thanks! I just pushed my changes!

seanmonstar (approved):

PR #67, opened 2021-12-13
Make body data types generic

stammw:
Get closer to the original design proposal by making the API generic over send and receive `Buf` data types.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 use std::time::SystemTime;
 
-use futures::{future, FutureExt};
+use futures::future;
 use h3_quinn::quinn;
 use rustls::{self, client::ServerCertVerified};
 use rustls::{Certificate, ServerName};
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -118,9 +118,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         eprintln!("Response: {:?} {}", resp.version(), resp.status());
         eprintln!("Headers: {:#?}", resp.headers());
 
-        while let Some(chunk) = stream.recv_data().await? {
+        while let Some(mut chunk) = stream.recv_data().await? {
             let mut out = tokio::io::stdout();
-            out.write_all(&chunk).await.expect("write_all");
+            out.write_all_buf(&mut chunk).await.expect("write_all");
             out.flush().await.expect("flush");
         }
         Ok::<_, Box<dyn std::error::Error>>(())
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -1,4 +1,4 @@
-use bytes::{Bytes, BytesMut};
+use bytes::{Buf, Bytes, BytesMut};
 use futures::future;
 use http::{request, HeaderMap, Response};
 use std::{
@@ -288,7 +288,7 @@ where
         Ok(resp)
     }
 
-    pub async fn recv_data(&mut self) -> Result<Option<Bytes>, Error> {
+    pub async fn recv_data(&mut self) -> Result<Option<impl Buf>, Error> {
         self.inner.recv_data().await
     }
 
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -295,7 +295,7 @@ where
     pub async fn recv_trailers(&mut self) -> Result<Option<HeaderMap>, Error> {
         let res = self.inner.recv_trailers().await;
         if let Err(ref e) = res {
-            if let crate::error::Kind::HeaderTooBig { .. } = e.kind() {
+            if e.is_header_too_big() {
                 self.inner.stream.stop_sending(Code::H3_REQUEST_CANCELLED);
             }
         }
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -1,12 +1,14 @@
-use bytes::{Buf, Bytes, BytesMut};
-use futures::future;
-use http::{request, HeaderMap, Response};
 use std::{
     convert::TryFrom,
     marker::PhantomData,
     sync::{atomic::AtomicUsize, Arc},
     task::{Context, Poll, Waker},
 };
+
+use bytes::{Buf, Bytes, BytesMut};
+use futures::future;
+use http::{request, HeaderMap, Response};
+use tracing::{trace, warn};
 
 use crate::{
     connection::{self, ConnectionInner, ConnectionState, SharedStateRef},
@@ -15,13 +17,12 @@ use crate::{
     proto::{frame::Frame, headers::Header, varint::VarInt},
     qpack, quic, stream,
 };
-use tracing::{trace, warn};
 
-pub fn builder<C: quic::Connection<Bytes>>() -> Builder<C> {
+pub fn builder() -> Builder {
     Builder::new()
 }
 
-pub async fn new<C, O>(conn: C) -> Result<(Connection<C>, SendRequest<C::OpenStreams>), Error>
+pub async fn new<C, O>(conn: C) -> Result<(Connection<C, Bytes>, SendRequest<O, Bytes>), Error>
 where
     C: quic::Connection<Bytes, OpenStreams = O>,
     O: quic::OpenStreams<Bytes>,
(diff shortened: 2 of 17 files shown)


h3/src/client.rs
```diff
@@ -1,51 +1,58 @@
...
+    T: quic::OpenStreams<B>,
+    B: Buf,
+{
     open: T,
     conn_state: SharedStateRef,
     max_field_section_size: u64, // maximum size for a header we receive
     // counts instances of SendRequest to close the connection when the last is dropped.
     sender_count: Arc<AtomicUsize>,
     conn_waker: Option<Waker>,
+    _buf: PhantomData<B>,
```
seanmonstar, line 44: Since this type never _holds_ a `B`, but instead just really _receives_ one and passes it on, you may want this to be `PhantomData<fn(B)>`. It's subtly different, and it alters how much auto traits (like `Send` and `Sync`) of `B` affect `SendRequest<T, B>`.

h3/src/client.rs
```diff
@@ -107,13 +116,15 @@ where
             max_field_section_size: self.max_field_section_size,
             sender_count: self.sender_count.clone(),
             conn_waker: self.conn_waker.clone(),
+            _buf: PhantomData::default(),
```
seanmonstar, line 119:
```suggestion
            _buf: PhantomData,
```

h3/src/frame.rs
```diff
@@ -2,33 +2,38 @@ use std::task::{Context, Poll};
...
+    stream::WriteBuf,
 };
 
 pub struct FrameStream<S>
 where
     S: RecvStream,
 {
     stream: S,
-    bufs: BufList<S::Buf>,
+    bufs: BufList<Bytes>,
```
seanmonstar, line 21: Just curious, why did this change from `S::Buf` to `Bytes`?

h3/src/proto/frame.rs
```diff
@@ -127,6 +184,66 @@ impl fmt::Display for Frame {
...
     }
 }
+
+impl fmt::Debug for Frame<PayloadLen> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        <Self as fmt::Display>::fmt(self, f)
+    }
+}
+
+impl<B> fmt::Display for Frame<B>
```
seanmonstar, line 194: I realize this was here before, so feel free to ignore, but: is there a need for `Display` for this type? It's an internal type, and the only fmt purpose is for logs, right? Since `Display` is often meant for "turn this type into a value you'd show a user", perhaps `Debug` is all that is needed.

h3/src/stream.rs
```diff
@@ -10,27 +10,152 @@ use crate::{
...
+            payload.chunk()
+        } else {
+            &[]
+        }
+    }
+
+    fn advance(&mut self, cnt: usize) {
+        let remaining = self.len - self.pos;
+        if remaining > 0 {
+            assert!(cnt <= remaining);
```
seanmonstar, line 151: I think this means that we can only advance the frame header first, and then a second call is needed to advance the payload. This probably won't happen unless `chunks_vectored` is implemented, but I'd recommend it, assuming the underlying QUIC library can use writev (`send_mmsg`, etc).

h3/src/stream.rs
```diff
@@ -49,7 +174,7 @@ where
     stream: S,
     ty: Option<StreamType>,
     push_id: Option<u64>,
-    buf: BufList<S::Buf>,
+    buf: BufList<Bytes>,
```
seanmonstar, line 177: Also curious why this change.

h3/src/quic.rs
```diff
@@ -110,7 +112,7 @@ pub trait SendStream<B: Buf> {
     fn poll_ready(&mut self, cx: &mut task::Context<'_>) -> Poll<Result<(), Self::Error>>;
 
     /// Send more data on the stream.
-    fn send_data(&mut self, data: B) -> Result<(), Self::Error>;
+    fn send_data<T: Into<WriteBuf<B>>>(&mut self, data: T) -> Result<(), Self::Error>;
```
seanmonstar, line 115: What does someone have to do to get this `Into<WriteBuf<B>>` to work? Just `impl Buf`, and it works automatically?

seanmonstar (commented):
Overall excellent work, thanks for all this effort ❤️

h3/src/frame.rs
stammw: In order to extract the next data chunk without consuming what's following it (either another `Data` header or the trailers) in the cheapest way possible, I chose to use `Bytes::split_to()`. As seen in :

```rust
impl BufList<Bytes> {
    pub fn take_chunk(&mut self, max_len: usize) -> Option<Bytes> {
        let chunk = self
            .bufs
            .front_mut()
            .map(|chunk| chunk.split_to(usize::min(max_len, chunk.remaining())));
            // [...]
    }
    // [...]
}
```

That would be harder or less efficient using only `Buf`. (I guess ?)

As we will call `<S::Buf as Buf>::copy_to_bytes()` at some point to return an owned "smart reference counted zero-copy slice" to the user, the best is to convert `RecvStream::Buf` into `Bytes` right after it's yeilded from the `quic` impl. So there are more chances it is a continuous chunk of bytes, and thus is cheaply convertible to `Bytes`.

h3/src/stream.rs
stammw: As argued in the last related comment, since `FrameStream` now stores `Bytes` instead of `S::Buf`, we'd need to map all the `S::Buf` into `Bytes` when constructing it via `FrameStream::with_bufs()`. To save up this iteration, best thing is to call `Buf::copy_to_bytes()` right after get our hands on each chunk.

h3/src/quic.rs
stammw: `WriteBuf` contains an array that will receive the encoded header. While it is most of the time a short slice of bytes, the fact that `VarInts` are used make the maximum possible size of this buffer big enough to slow down method calls that take the `Into<WriteBuf>` by value.  QUIC impls might want to store each data in a collection so it's best to take ownership of them (I guess ?).

Passing `Into<WriteBuf>` allows us to move only the wrapped frame, and allocate the encoding buffer on the stack only when it is necessary. 

But maybe it is possible to do something like `Into<impl Buf>` ?

examples/client.rs
```diff
@@ -118,9 +118,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         eprintln!("Response: {:?} {}", resp.version(), resp.status());
         eprintln!("Headers: {:#?}", resp.headers());
 
-        while let Some(chunk) = stream.recv_data().await? {
+        while let Some(mut chunk) = stream.recv_data().await? {
             let mut out = tokio::io::stdout();
-            out.write_all(&chunk).await.expect("write_all");
+            out.write_buf(&mut chunk).await.expect("write_all");
```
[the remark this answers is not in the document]
stammw: Thanks! Done in original commit.

h3/src/client.rs
stammw: Expert things! Yay ! If you can think of a link where I could learn more about this I'd be interested. Thanks !

Done in original commit.

h3/src/proto/frame.rs
stammw: Done in 2bb75d6.

Don't know why I implemented `Display` :sweat_smile: .

h3/src/stream.rs
stammw: Goo catch ! I think it was actually a bug. :sweat_smile:. Thanks !

Done in 0a6e000

h3/src/client.rs
seanmonstar: This comes from:

- [`std::marker::PhantomData`](https://doc.rust-lang.org/std/marker/struct.PhantomData.html#ownership-and-the-drop-check)
- [Nomicon](https://doc.rust-lang.org/nomicon/phantom-data.html)
- Experience and playing in the playground.

It'd be cool if there was a clearer resource that wasn't quite so deep in type theory...

stammw: Thanks for sharing ! I should spend more time reading the Nomicon.

h3/src/quic.rs
stammw: Are you ok with this justification or do you think it's not acceptable / could be a lot better?

seanmonstar: Yea, the justification seems fine. I was more wondering what does a user of this library write in order to make this work. I didn't examine the implementations closely enough, is it sufficient to just pass anything that implements `Buf`? Or do they need to manually do `impl Into<WriteBuf<MyBuf>> for MyBuf`?

stammw: Quic implementers have to call `Into<WriteBuf>` when they want to send the frame on the wire. Then they'll get a `Buf` impl. A nice thing here is that they can buffer `Into<WriteBuf>` into a list cheaply, while still being able to get the size of each frame. 

WriteBuf implementation is the role of h3.

seanmonstar (approved):

PR #68, opened 2021-12-13
WIP: Fix/stammw/clippy fixes

stammw:
fixes #59

Issue #65, opened 2021-11-30
Implement GoAway


stammw:
Add the graceful shutdown in conformity with [5.2 Connection Shutdown](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-5.2).

Take inspiration in `h2` and `hyper` to offer a usable API for this.

PR #72, opened 2021-12-25
Graceful shutdown implementation

stammw:
Sends `GO_AWAY` frames and tracks ongoing requests in server. Then make `accept()` return `Ok(None)` when all requests have completed, so the user will definitively close by dropping `Connection`.

Adresses #65. Based on #67 and #68.

### DIFF
--- a/h3/src/quic.rs
+++ b/h3/src/quic.rs
@@ -7,7 +7,7 @@ use std::task::{self, Poll};
 
 use bytes::Buf;
 
-pub use crate::proto::stream::StreamId;
+pub use crate::proto::stream::{InvalidStreamId, StreamId};
 pub use crate::stream::WriteBuf;
 
 // Unresolved questions:
--- a/h3/src/quic.rs
+++ b/h3/src/quic.rs
@@ -7,6 +7,7 @@ use std::task::{self, Poll};
 
 use bytes::Buf;
 
+pub use crate::proto::stream::StreamId;
 pub use crate::stream::WriteBuf;
 
 // Unresolved questions:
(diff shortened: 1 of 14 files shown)


seanmonstar:
This also could use a rebase, so it's easier to review just the related changes.

seanmonstar:
Another rebase would help remove all the clippy changes that are sprinkled about 🙏

stammw:
Ho no ! I really should be a bit more cautious on this. (Done)

h3/src/quic.rs
```diff
@@ -121,7 +122,7 @@ pub trait SendStream<B: Buf> {
     fn reset(&mut self, reset_code: u64);
 
     /// Get QUIC send stream id
-    fn id(&self) -> u64;
+    fn id(&self) -> StreamId;
```
seanmonstar, line 125: This is the public trait that people implement to integrate a QUIC library... Should we require the stronger type here? It seems we could continue to just allow it to be `u64`, and convert internally. Unless there's a strong reason for people to use it the type themselves, being more conservative may be advantageous.

stammw: I made `StreamId` constructible only via `TryFrom` in a new commit. But I'm not sure if we shouldn't rather `panic!` when an implementer tries to build an invalid StreamId.

seanmonstar: Yea, I don't think the trait should have `fn id(&self) -> Result<..>`. The ID of the stream should available easily and immediately. What does it mean for a stream to not be able to tell `h3` its valid ID?

stammw: In  `quic::SendStream::stream_id()`, it means that the quic implementation isn't compliant. So I made it panic there.

stammw:
@seanmonstar Can you have another look at this, please ?

seanmonstar (approved):

PR #73, opened 2022-01-06
File server example

stammw:
Having a server example that only return `404` can be disappointing for people looking into the project.

This PR makes it a file server, taking a root directory in it's arguments.


examples/server.rs
```diff
@@ -113,31 +90,52 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+) -> Result<(), Box<dyn std::error::Error>>
 where
     T: BidiStream<Bytes>,
 {
-    let resp = http::Response::builder()
-        .status(http::StatusCode::NOT_FOUND)
-        .body(())
-        .unwrap();
+    let (status, to_serve) = match serve_root.as_deref() {
+        None => (StatusCode::IM_A_TEAPOT, None),
```
seanmonstar, line 124: Does a browser or client know what to do with such a response?

examples/server.rs
```diff
@@ -1,55 +1,52 @@
...
-    pub key: PathBuf,
-
-    #[structopt(long, default_value = "4433")]
-    pub port: u16,
+    #[structopt(
+        long,
+        short,
+        help = "Certificate for TLS. \
+                If present, `--key` is mandatory. \
+                If ommitted, a selfsigned certificate will be generated."
```
seanmonstar, line 44:
```suggestion
                If omitted, a selfsigned certificate will be generated."
```

examples/server.rs
```diff
@@ -1,55 +1,52 @@
...
-    pub debug: bool,
-
-    #[structopt(long, default_value = "4433")]
-    pub port: u16,
+    #[structopt(
+        name = "dir",
+        short,
+        long,
+        help = "Root directory of the files to serve. \
+                If ommited, server will transform into a teapot." // We're in IOT era!
```
seanmonstar, line 21:
```suggestion
                If omited, server will transform into a teapot." // We're in IOT era!
```

examples/server.rs
stammw: I don't think so, I felt like putting a joke in there when miss-configured (root directory is missing). Unfortunately, browsers don't get the humor: 

image

I didn't want to automatically open the current directory to the world, so I made the argument optional and said that the server had a teapot feature instead.

If you want to get it removed, I can either make the parameter mandatory or set it to a mostly sane default, like `www` for example.

seanmonstar: I agree it shouldn't default to the current directory. Since this is an example, I think it'd be good if a simple "hello world" response were sent with a 200 status.

seanmonstar (approved):
LGTM! I'd recommend the change from 418 to a 200 "hello world" by default, but the rest is great.

Issue #55, opened 2021-09-30
Drop AsyncWrite usage in h3-quinn once GATs land


stammw:
We were forced to use `AsyncWrite` in `h3_quinn::SendStream` because Quinn only offers lifetime-bound Futures to poll for writes: 
`Write<'a, S>`.

Generic associated type could cancel this limitation, allowing to store them into `h3_quinn::SendStream`.

This will have positive implication for performance and will remove the necessity to `Pin` the stream and to downcast the error type from `IOError` to `quinn::WriteError`.

Ralith:
quinn could also be modified to expose raw `poll_*` style methods alongside important futures, if a better approach is required in the mean time.

stammw:
I remember that I thought of using the `poll_*` methods, but can't recall why I didn't ask for them to be made available. 
I guess this solution will cancel the need of GATs whatsoever.

inflation:
This should be closed after quinn 0.9.0 upgrade.

PR #80, opened 2022-02-09
quinn main update using async traits

FrankSpitulski:
Quinn has removed the Open* Future structs that h3 was storing for polling. I tried storing it myself, but the lifetimes got very complicated. Then I tried using async traits and removing polling but got stuck again with lifetimes. I'll share my work in case this is the path you want to go down for keeping up to date with quinn, but in any case a breakage is coming with 0.9.0.

### DIFF
--- a/h3-quinn/Cargo.toml
+++ b/h3-quinn/Cargo.toml
@@ -7,8 +7,8 @@ edition = "2018"
 [dependencies]
 h3 = { path = "../h3" }
 bytes = "1"
-quinn = "0.8.0"
-quinn-proto = "0.8.0"
+quinn = { git = "https://www.github.com/quinn-rs/quinn", branch = "main" }
+quinn-proto = { git = "https://www.github.com/quinn-rs/quinn", branch = "main" }
 futures = "0.3"
 
 [dev-dependencies]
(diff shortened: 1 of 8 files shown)


seanmonstar:
Is the short of it that quinn is removing polling methods and only exposing `async fn`s for everything, now?

FrankSpitulski:
There are still some polling methods left. @Ralith or @djc may be able to answer about future plans.

FrankSpitulski:
Some discussion is in here https://gitter.im/djc/quinn

Ralith:
See also previous discussion in https://github.com/hyperium/h3/issues/55.

> Is the short of it that quinn is removing polling methods and only exposing async fns for everything, now?

That's the current state; it's much more maintainable. We could offer polling methods, but doing so is not a trivial amount of work.

seanmonstar:
I very much understand that. Encapsulating them in a trait is still annoying, but exposing them is too.

DemiMarie:
GATs are now stable.

djc:
We just merged https://github.com/quinn-rs/quinn/pull/1357.

FrankSpitulski:
I'll close this since it appears quinn is keeping named futures instead of async traits.

PR #81, opened 2022-02-10
Fix write to check poll_ready before sending data

### LINKED ISSUE #78: write method calls send_data before poll_ready
On [stream.rs#L28-L29](https://github.com/hyperium/h3/blob/master/h3/src/stream.rs#L28-L29), `stream.send_data` is called prior to `stream.poll_ready`. 

`stream.poll_ready` should be called first, similar to as described in https://docs.rs/futures/latest/futures/sink/trait.Sink.html#tymethod.poll_ready. 

Thanks for considering this issue, and let me know if you have any questions!

seanmonstar:
Closes #78

### DIFF
--- a/h3/src/stream.rs
+++ b/h3/src/stream.rs
@@ -25,8 +25,8 @@ where
     D: Into<WriteBuf<B>>,
     B: Buf,
 {
-    stream.send_data(data)?;
     future::poll_fn(|cx| stream.poll_ready(cx)).await?;
+    stream.send_data(data)?;
 
     Ok(())
 }


stammw (approved):
Thanks!

stammw:
Oops, approved this a bit too quickly... I looks like this swap makes all the transfers fail. I vaguely remember having this problem. @camshaft suggestion might help.

Also, do you think the transfer should be initiated at the first `send_data()` call ?

stammw:
Right, maybe I should have read the issue beforehand :sweat_smile: . Thanks!

While we're at it, we might want to ditch `AsyncWrite` by doing what Ralith suggested [in this issue](https://github.com/hyperium/h3/issues/55#issuecomment-1003624209):

> quinn could also be modified to expose raw `poll_*` style methods alongside important futures, if a better approach is required in the mean time. That's assuming it's not practical to modify h3 to push the pinning/lifetime issues downstream.

Or maybe implement `Sink`directly in `quinn` and use it.

camshaft:
Yeah AsyncWrite isn't the most efficient way of transferring data, especially when the data is already a `Bytes`. With [s2n-quic](https://docs.rs/s2n-quic/latest/s2n_quic/stream/struct.SendStream.html) we've exported poll functions for all of the stream operations. It leads to a really [clean implementation](https://github.com/aws/s2n-quic/blob/a19457fe64d32f1b3b86c6270864cddca49ea716/quic/s2n-quic-h3/src/s2n_quic.rs#L356) of the h3 traits.

stammw:
Nice ! I didn't know about this project.

Issue #53, opened 2021-09-30
Make header decoding fail when MAX_FIELD_SECTION_SIZE is exceeded


stammw:
Header list size constraints are currently implemented in conformity with [10.5.1. Limits on Field Section Size](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#name-limits-on-field-section-siz), but the whole header is decoded before the error is thrown.

It would make more sense if the error was thrown as soon as the header is considered too big.

Modify `qpack::decode_stateless()` so it fails as soon as the limit is exceeded.

PR #84, opened 2022-03-03
Make header decoding fail when MAX_FIELD_SECTION_SIZE is exceeded

Ruben2424:
Hi,

I solved #53. I am new to Rust, so every feedback is appreciated.

### DIFF
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -289,9 +289,8 @@ where
                 Code::H3_GENERAL_PROTOCOL_ERROR.with_reason("Did not receive response headers")
             })?;
 
-        let decoded;
-        if let Frame::Headers(ref mut encoded) = frame {
-            decoded = match qpack::decode_stateless(encoded, self.inner.max_field_section_size) {
+        let decoded = if let Frame::Headers(ref mut encoded) = frame {
+            match qpack::decode_stateless(encoded, self.inner.max_field_section_size) {
                 Err(qpack::DecoderError::HeaderTooLong(cancel_size)) => {
                     self.inner.stop_sending(Code::H3_REQUEST_CANCELLED);
                     return Err(Error::header_too_big(
@@ -301,7 +300,7 @@ where
                 }
                 Ok(decoded) => decoded,
                 Err(e) => return Err(e.into()),
-            };
+            }
         } else {
             return Err(
                 Code::H3_FRAME_UNEXPECTED.with_reason("First response frame is not headers")
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -288,9 +288,19 @@ where
             .ok_or_else(|| {
                 Code::H3_GENERAL_PROTOCOL_ERROR.with_reason("Did not receive response headers")
             })?;
-
-        let decoded = if let Frame::Headers(ref mut encoded) = frame {
-            qpack::decode_stateless(encoded)?
+        
+        let decoded;
+        if let Frame::Headers(ref mut encoded) = frame {
+            decoded= match qpack::decode_stateless(encoded,self.inner.max_field_section_size){
+                Err(qpack::DecoderError::HeaderToLong(cancel_size)) => {
+                    self.inner.stop_sending(Code::H3_REQUEST_CANCELLED);
+                    return Err(Error::header_too_big(
+                        cancel_size,
+                        self.inner.max_field_section_size,
+                    ))},
+                Ok(decoded) => decoded,
+                Err(e) => return Err(e.into()),
+            };
         } else {
             return Err(
                 Code::H3_FRAME_UNEXPECTED.with_reason("First response frame is not headers")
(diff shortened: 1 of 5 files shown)


h3/src/client.rs
```diff
@@ -288,27 +288,29 @@ where
             .ok_or_else(|| {
                 Code::H3_GENERAL_PROTOCOL_ERROR.with_reason("Did not receive response headers")
             })?;
-
-        let decoded = if let Frame::Headers(ref mut encoded) = frame {
-            qpack::decode_stateless(encoded)?
+        
+        let decoded;
```
seanmonstar, line 292: To make this lint happy, you can write `let decoded = if let Frame::Headers(..)`.

h3/src/connection.rs
```diff
@@ -340,6 +340,7 @@ where
 
     /// Receive trailers
     pub async fn recv_trailers(&mut self) -> Result<Option<HeaderMap>, Error> {
+
```
seanmonstar, line 343:
```suggestion
```

h3/src/connection.rs
```diff
@@ -366,11 +367,12 @@ where
         }
 
         let qpack::Decoded {
-            fields, mem_size, ..
-        } = qpack::decode_stateless(&mut trailers)?;
-        if mem_size > self.max_field_section_size {
-            return Err(Error::header_too_big(mem_size, self.max_field_section_size));
-        }
+            fields, ..
+        } =  match qpack::decode_stateless(&mut trailers,self.max_field_section_size){
```
seanmonstar, line 371:
```suggestion
        } = match qpack::decode_stateless(&mut trailers,self.max_field_section_size) {
```

h3/src/qpack/decoder.rs
```diff
@@ -35,6 +35,7 @@ pub enum Error {
     MissingRefs(usize),
     BadBaseIndex(isize),
     UnexpectedEnd,
+    HeaderToLong(u64),
```
seanmonstar, line 38:
```suggestion
    HeaderTooLong(u64),
```

h3/src/qpack/decoder.rs
```diff
@@ -51,6 +52,7 @@ impl std::fmt::Display for Error {
             Error::MissingRefs(n) => write!(f, "missing {} refs to decode bloc", n),
             Error::BadBaseIndex(i) => write!(f, "out of bounds base index: {}", i),
             Error::UnexpectedEnd => write!(f, "unexpected end"),
+            Error::HeaderToLong(_) => write!(f, "header to long "),
```
seanmonstar, line 55:
```suggestion
            Error::HeaderTooLong(_) => write!(f, "header too long"),
```

h3/src/qpack/decoder.rs
```diff
@@ -235,6 +237,10 @@ pub fn decode_stateless<T: Buf>(buf: &mut T) -> Result<Decoded, Error> {
             _ => return Err(Error::UnknownPrefix(buf.chunk()[0])),
         };
         mem_size += field.mem_size() as u64;
+        // Cancel decoding if the header is considered too big
+        if mem_size > max_size{
```
seanmonstar, line 241:
```suggestion
        if mem_size > max_size {
```

seanmonstar (changes requested):
Thanks for the PR! I left some style comments inline. I'm also curious, is there a test case we could add to verify the change?

h3/src/client.rs
Ruben2424: I tried this, but then `decoded` has the type of `()` which leads to a type mismatch in this line `let qpack::Decoded { fields, .. } = decoded;`. This is the code:

```rust
        let decoded = if let Frame::Headers(ref mut encoded) = frame {
            match qpack::decode_stateless(encoded, self.inner.max_field_section_size) {
                Err(qpack::DecoderError::HeaderTooLong(cancel_size)) => {
                    self.inner.stop_sending(Code::H3_REQUEST_CANCELLED);
                    return Err(Error::header_too_big(
                        cancel_size,
                        self.inner.max_field_section_size,
                    ));
                }
                Ok(decoded) => decoded,
                Err(e) => return Err(e.into()),
            };
        } else {
            return Err(
                Code::H3_FRAME_UNEXPECTED.with_reason("First response frame is not headers")
            );
        };

        let qpack::Decoded { fields, .. } = decoded;
```
What did i wrong?

Ruben2424:
I fixed the style issues and added a test. I actually didn't know the cargo fmt tool.

h3/src/client.rs
seanmonstar: I think you just need to the remove the semi-colon from the `match` expression.

Ruben2424: Ah, that helped. Thank you.

seanmonstar (approved):
Looks good here, thanks!

PR #85, opened 2022-03-06
Do not immediately close QPACK decoder and encoder streams

g2p:
The logic for locating the control stream accepted all pending streams,
took ownership, and immediately dropped everything but the control
stream.  The streams are closed on drop.

Closing critical QPACK streams will trigger an H3_CLOSED_CRITICAL_STREAM
error on the other end and scupper the connection:
https://quicwg.org/base-drafts/draft-ietf-quic-qpack.html#name-encoder-and-decoder-streams

To fix this, transfer ownership of the two accepted QPACK streams to the
connection object.  This doesn't do anything to make dynamic QPACK work,
but it seems to be enough to interoperate with at least neqo and firefox.

### DIFF
--- a/h3/src/connection.rs
+++ b/h3/src/connection.rs
@@ -80,6 +80,8 @@ where
     conn: C,
     control_send: C::SendStream,
     control_recv: Option<FrameStream<C::RecvStream>>,
+    decoder_recv: Option<AcceptedRecvStream<C::RecvStream>>,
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Nice catch!

g2p:
(rebased so rustls works in CI)

PR #86, opened 2022-03-06
Dependency updates

g2p:
Trivial updates


seanmonstar:
Hrm, looks like rustls now requires `resolver` in Cargo, so I guess we have to increase the MSRV from 1.50...

g2p:
Ah, sure! Convenient that h3 is unreleased.

seanmonstar (approved):

PR #87, opened 2022-03-06
Dependency reduction

g2p:
These changes reduce build times for users of these crates.


seanmonstar:
Awesome work, thank you so much :D

Issue #28, opened 2020-11-30
Send GREASE


seanmonstar:
The spec recommends ("SHOULD") exercising peer implementations to ensure that they can handle new features when they come out. This is done by sending random settings, stream and frame types (that fit the `0x1f * N - 0x21` scheme). We should do that by default, with an option to disable it.

- Sending a GREASE setting
- Opening a GREASE stream type
- Sending a GREASE frame on every stream

Ruben2424:
I didn't found this in the newest spec. Is this still relevant?

seanmonstar:
[From the spec regarding SETTINGS](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-7.2.4.1):

> Setting identifiers of the format `0x1f * N + 0x21` for non-negative integer values of N are reserved to exercise the requirement that unknown identifiers be ignored. Such settings have no defined meaning. Endpoints SHOULD include at least one such setting in their SETTINGS frame.

Regarding [stream types](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-6.2.3):

> Stream types of the format `0x1f * N + 0x21` for non-negative integer values of N are reserved to exercise the requirement that unknown types be ignored. These streams have no semantics, and can be sent when application-layer padding is desired. They MAY also be sent on connections where no data is currently being transferred.

And [frame types](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#section-7.2.8):

> Frame types of the format `0x1f * N + 0x21` for non-negative integer values of N are reserved to exercise the requirement that unknown types be ignored ([Section 9](https://quicwg.org/base-drafts/draft-ietf-quic-http.html#extensions)). These frames have no semantics, and MAY be sent on any stream where frames are allowed to be sent.

-----

So only the note about reserved SETTINGS includes the language that an implementation SHOULD send these. The others don't have such strong language. I noticed that https://github.com/cloudflare/quiche includes a `grease` option that will try to send all 3 of these.

Ruben2424:
Ok, I will try to implement that.

PR #92, opened 2022-06-09
Grease Setting

Ruben2424:
Hi,

i tried to solve #28. I think the settings grease works.
But when should the frame and stream grease be sent?

### DIFF
--- a/h3/Cargo.toml
+++ b/h3/Cargo.toml
@@ -14,6 +14,7 @@ futures-util = { version = "0.3", default-features = false }
 http = "0.2.3"
 tokio = { version = "1", features = ["sync"] }
 tracing = "0.1.18"
+rand = "0.8.5"
 
 [dev-dependencies]
 assert_matches= "1.3.0"
--- a/h3/Cargo.toml
+++ b/h3/Cargo.toml
@@ -14,7 +14,7 @@ futures-util = { version = "0.3", default-features = false }
 http = "0.2.3"
 tokio = { version = "1", features = ["sync"] }
 tracing = "0.1.18"
-rand = "0.8.5"
+fastrand = "1.7.0"
 
 [dev-dependencies]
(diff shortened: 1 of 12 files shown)


seanmonstar (commented):
- [Stream Grease](https://httpwg.org/specs/rfc9114.html#stream-grease) suggests to do it on an idle connection. [quiche](https://github.com/cloudflare/quiche/blob/4bda0917dd5aa535f39214063ee85c2cad00ceb2/quiche/src/h3/mod.rs#L906) does it immediately.
- [Frame Grease](https://httpwg.org/specs/rfc9114.html#frame-reserved) doesn't give a suggestion anymore, just that they *MAY* be sent on any stream. [quiche](https://github.com/cloudflare/quiche/blob/4bda0917dd5aa535f39214063ee85c2cad00ceb2/quiche/src/h3/mod.rs#L1054) seems to send a grease frame on the first request stream, as the first frame. It currently only does this once per connection.

We don't need to do it in exactly the same order as quiche does, in fact being slightly different is probably a good thing for the purpose of grease.

seanmonstar:
Thanks so much for starting on this! I forgot to mention that in the review, but this is awesome <3

Ruben2424:
Do you have any idea how to check whether a connection is idle? And when should the check happen?

seanmonstar:
I've restarted the CI jobs a few times, thinking there was something racey, but now I'm wondering if they're noticing a difference in streams/frames received, since it seems they're tests expecting the connection to close in a certain way?

Ruben2424:
Ok, i will see if i find the Problem. On my local machine the tests are successful with `rustc 1.62.1 `.

Ruben2424:
ah now the tests are successful

seanmonstar (approved):
Awesome work!!

PR #94, opened 2022-06-17
move generic Buffer parameter to fn

yu-re-ka:


yu-re-ka:
I still think this is the right way to do it, making both the code and the api cleaner.
But please let me know your opinion so that I can either close or rebase this.

PR #96, opened 2022-06-18
recv_data always returns Bytes

yu-re-ka:
frame.rs and buf.rs are hardcoded to providing Bytes anyways.

This is useful when writing the result to a hyper::Body channel since that only takes Bytes.

### DIFF
--- a/h3/src/connection.rs
+++ b/h3/src/connection.rs
@@ -337,7 +337,7 @@ where
     S: quic::RecvStream,
 {
     /// Receive some of the request body.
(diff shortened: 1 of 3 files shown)


seanmonstar:
I think the point was that we might change the return type at some point... it's not the best solution, but it is possible to use `Buf::copy_to_bytes` to materialize a `Bytes` (and if possible, the `impl Buf` will make that operation cheap).

yu-re-ka:
I see, thanks!

Issue #83, opened 2022-03-02
Provide a way to split a RequestStream into send and receive halves


g2p:
RequestStream should implement the BidiStream trait; a split version will be able to stream request bodies and response data from different tasks without angering the borrow checker.

seanmonstar:
I think we accounted for this with the `BidiStream` trait, and we just need to implement it for the `client::RequestStream` and `server::RequestStream` types, right?

PR #97, opened 2022-06-18
splittable RequestStream

yu-re-ka:
#83

### DIFF
--- a/h3/src/frame.rs
+++ b/h3/src/frame.rs
@@ -1,3 +1,4 @@
+use std::marker::PhantomData;
 use std::task::{Context, Poll};
 
 use bytes::{Buf, Bytes};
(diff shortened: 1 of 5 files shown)


h3/src/quic.rs
```diff
@@ -32,7 +32,7 @@ impl<'a, E: Error + 'a> From<E> for Box<dyn Error + 'a> {
 /// Trait representing a QUIC connection.
 pub trait Connection<B: Buf> {
     /// The type produced by `poll_accept_bidi()`
-    type BidiStream: SendStream<B> + RecvStream;
+    type BidiStream: BidiStream<B>;
```
seanmonstar, line 35: If we keep the previous bounds, then implementors aren't _required_ to provide splittable streams, just send and receive. If they implement BidiStream, as an opt-in, then the streams can be splittable.

But by not requiring it, we allow QUIC implementations that don't have an internal cheap way to split.

yu-re-ka:
I redesigned the split mechanism with that use case in mind, please have another look

h3/src/frame.rs
```diff
@@ -12,26 +13,21 @@ use crate::{
...
-    S: RecvStream,
-{
+pub struct FrameStream<S, B> {
     stream: S,
     bufs: BufList<Bytes>,
     decoder: FrameDecoder,
     remaining_data: usize,
     /// Set to true when `stream` reaches the end.
     is_eos: bool,
+    _phantom_buffer: PhantomData<B>,
```
seanmonstar, line 27: Was this marker required? How come? I imagine probably yes, but I'm struggling to catch exactly where/why...

yu-re-ka:
```
impl<S, B> FrameStream<S, B>
where
    S: BidiStream<B>,
    B: Buf,
{
```

If FrameStream doesn't have the B parameter, this gives an error that B is unconstrained...

yu-re-ka: #94 would solve this

seanmonstar (approved):

Issue #103, opened 2022-07-10
Duplicate headers are lost


yu-re-ka:
When a response has multiple set-cookie headers, the client only receives one.

Minimal example to reproduce this:

```diff
diff --git a/examples/server.rs b/examples/server.rs
index 6a0d90f..1cb7add 100644
--- a/examples/server.rs
+++ b/examples/server.rs
@@ -135,7 +135,11 @@ where
         }
     };
 
-    let resp = http::Response::builder().status(status).body(()).unwrap();
+    let resp = http::Response::builder()
+        .header("Set-Cookie", "foo=foo")
+        .header("Set-Cookie", "bar=bar")
+        .status(status)
+        .body(()).unwrap();
 
     match stream.send_response(resp).await {
         Ok(_) => {

```
```
$ cargo run --example server -- --listen=127.0.0.1:4433 &
$ cargo run --example client -- https://127.0.0.1:4433 --insecure=true
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/examples/client 'https://127.0.0.1:4433' --insecure=true`
DNS Lookup for https://127.0.0.1:4433/: 127.0.0.1:4433
QUIC connected ...
Sending request ...
Receiving response ...
Response: HTTP/3.0 200 OK
Headers: {
    "set-cookie": "foo=foo",
}

$

seanmonstar:
Fixed in #104.

PR #104, opened 2022-07-10
preserve duplicate headers

yu-re-ka:
#103


seanmonstar:
Nice catch! Think it'd be possible to include a unit test so we don't mess this up later?

yu-re-ka:
End-to-end or just for the HeaderIter?

yu-re-ka:
Added something, let be know if this is what you wanted

seanmonstar (approved):
Excellent, thank you!

PR #105, opened 2022-07-18
Browser example

Ruben2424:
Hi,
i added an example to test the Server against the chromium Browser.


seanmonstar (approved):
Sounds good to me!

PR #108, opened 2022-07-25
server docs

Ruben2424:
Hi,

i wrote some docs for the server module. I hope they are useful.
If there is anything missing or wrong, please tell me and I will fix it.

### DIFF
--- a/h3/src/server.rs
+++ b/h3/src/server.rs
@@ -1,5 +1,3 @@
-#![allow(rustdoc::private_intra_doc_links)]
-
 //! This module provides methods to create a http/3 Server.  
 //!
 //! It allows to accept incoming requests, and send responses.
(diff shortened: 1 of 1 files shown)


h3/src/server.rs
```diff
@@ -1,3 +1,41 @@
+#![allow(rustdoc::private_intra_doc_links)]
```
seanmonstar, line 1: Which link needs this setting? Perhaps it's better to just not have the link... Is this option stable?

Ruben2424: It is the link to `Builder::new()`.
How do I find out weather the option is stable?
```
warning: public documentation for `new` links to private item `Builder::new`
   --> h3/src/server.rs:102:86
    |
102 |     /// This method creates a new Connection for Servers with default settings see [`Builder::new()`].  
    |                                                                                      ^^^^^^^^^^^^^^ this item is private
    |
    = note: `#[warn(rustdoc::private_intra_doc_links)]` on by default
    = note: this link will resolve properly if you pass `--document-private-items`

warning: `h3` (lib doc) generated 1 warning
```

seanmonstar: I'd suggest that either, a) we document that you use `builder()` always, and don't refer to `Builder::new()`, or b) make `new` public and the preferred way to do it. I think I prefer (a), but could here reasoning for (b). Either way, we no longer would need to link to a private item.

seanmonstar (approved):

PR #114, opened 2022-08-09
Set up spec compliance checks

### LINKED ISSUE #109: Set up spec compliance checker
The s2n-quic maintainers created a neat [tool to check compliance](https://github.com/aws/s2n-quic/tree/main/common/duvet) with specifications with comments in the code. We can set it up to run as a CI job against [RFC 9114](https://httpwg.org/specs/rfc9114.html), and have it publish to the `gh-pages` branch.

This [GH Action](https://github.com/aws/s2n-quic/tree/main/.github/actions/duvet) is very close to what we want. The only difference is to remove the S3 bucket stuff, and just push to `gh-pages`.

Once this is setup, we can start systematically adding comments or fixing things to improve our handling of the HTTP/3 spec.

eagr:
close #109 

what will the action do
* clone aws/s2n-quic
* build and cache Duvet
* checkout `gh-pages` and auto-merge `master`
* extract checks from rfc specs
* generate report
* commit and push to origin/gh-pages

any pointers? @seanmonstar @camshaft

### DIFF
--- a/.github/actions/compliance/action.yml
+++ b/.github/actions/compliance/action.yml
@@ -1,0 +1,93 @@
+name: 'Spec compliance'
+description: 'Use Duvet to generate spec compliance report'
+
+inputs:
+  extract-script:
+    description: 'Path to script that extract spec checks'
+    required: true
+  report-script:
+    description: 'Path to script that generates a Duvet report'
+    required: true
+  dir:
+    description: 'Path to the directory where h3 is cloned'
+    default: ${{ github.workspace }}
+    required: false
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v3
+      with:
+        repository: aws/s2n-quic
+        path: s2n-quic
+        token: ${{ github.token }}
+
+    - uses: actions-rs/toolchain@v1.0.7
+      id: toolchain
+      with:
+        toolchain: stable
+        override: true
+
+    - uses: camshaft/rust-cache@v1
+
+    - name: Clean up cache
+      working-directory: ${{ inputs.dir }}
+      run: |
+        rm -f target/release/duvet
+        rm -f s2n-quic/common/duvet/target/release/duvet
+      shell: bash
+
+    - name: Cache duvet
+      uses: actions/cache@v3.0.3
+      continue-on-error: true
+      with:
+        path: ${{ inputs.dir }}/target/release/duvet
+        key: ${{ runner.os }}-${{ steps.toolchain.outputs.rustc_hash }}-${{ github.job }}-duvet-${{ hashFiles(format('{0}/Cargo.lock', inputs.dir)) }}-${{ hashFiles(format('{0}/s2n-quic/common/duvet/**', inputs.dir)) }}
+
+    - name: Install duvet
+      working-directory: ${{ inputs.dir }}/s2n-quic/common/duvet
+      run: |
+        if [ ! -f ../../../target/release/duvet ]; then
+          mkdir -p ../../../target/release
+          cargo build --release
+          cp target/release/duvet ../../../target/release/duvet
+        fi
+
+        echo "${{ inputs.dir }}/target/release" >> $GITHUB_PATH
+      shell: bash
+
+    - name: Checkout gh-pages
+      run: |
+        git config --local user.email "github-actions[bot]@users.noreply.github.com"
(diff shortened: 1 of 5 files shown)


eagr:
Could somehow *cache* the extracted checks so that we don't need to generate them on every run. But the extraction seems quite fast, almost instant. Your call.

.github/actions/compliance/action.yml
```diff
@@ -0,0 +1,93 @@
...
+        echo "${{ inputs.dir }}/target/release" >> $GITHUB_PATH
+      shell: bash
+
+    - name: Checkout gh-pages
+      run: |
+        git config --local user.email "github-actions[bot]@users.noreply.github.com"
+        git config --local user.name "github-actions[bot]"
+
+        if  git rev-parse --verify --quiet gh-pages; then
+          git checkout gh-pages
```
camshaft, line 65: it's usually better to do an orphan branch so you don't include all of the main source files

.github/actions/compliance/action.yml
```diff
@@ -0,0 +1,93 @@
...
+
+    - name: Commit
+      working-directory: ${{ inputs.dir }}
+      run: |
+        git add ci/compliance/*.html
+        git commit -m "${{ github.triggering_actor }}-${{ github.sha }}-${{ github.job }}#${{ github.run_number }}"
+      shell: bash
+
+    - name: Push to gh-pages
+      uses: ad-m/github-push-action@master
```
camshaft, line 89: does this only publish on push events? this workflow is also run on pull_request events so it will likely fail for users that don't have permission to push to gh-pages.

camshaft (commented):
Up to you on committing the extract step. The process is basically instant. We do in s2n-quic just to make it easier to ensure all of the requirements are processed correctly as part of developing duvet.

I do recommend committing the RFC text (like you've already done) so you're not having to request it every time it's executed.

.github/workflows/CI.yml
```diff
@@ -172,3 +172,19 @@ jobs:
         with:
           command: +nightly
           args: fuzz run fuzz_varint -- -runs=1
+
+  compliance:
```
seanmonstar, line 176: Should this be run only _after_ a merge? This currently will push to gh-pages on every PR push, I think?

eagr: Set the pushing step to only execute if the workflow is triggered by pushing to master

.github/actions/compliance/action.yml
eagr: Would this still be a problem?

seanmonstar (approved):
Phenomenal work!

seanmonstar:
And, it's now [rendered](https://hyper.rs/h3/ci/compliance/report.html)! ... Time for us to start filling in comments 😅

eagr:
I could help out with that.

Ruben2424:
> And, it's now [rendered](https://hyper.rs/h3/ci/compliance/report.html)! ... Time for us to start filling in comments sweat_smile

Are these the `//=` and `//#` comments?

seanmonstar:
I haven't verified the exact format, but yes, those, essentially.

camshaft:
That is the correct format. You can see tons of examples in the s2n-quic codebase:

https://github.com/aws/s2n-quic/blob/3fd1e4d30b42b2297703a438f5530cf2b98ceb14/quic/s2n-quic-core/src/recovery/cubic.rs#L79-L83

```rust
//= https://www.rfc-editor.org/rfc/rfc9002#section-7.3.2
//# If the congestion window is reduced immediately, a
//# single packet can be sent prior to reduction.  This speeds up loss
//# recovery if the data in the lost packet is retransmitted and is
//# similar to TCP as described in Section 5 of [RFC6675].
```

You can also annotate tests with `//= type=test` to track both implementation and the test of the implementation (see https://github.com/aws/s2n-quic/blob/3fd1e4d30b42b2297703a438f5530cf2b98ceb14/quic/s2n-quic-core/src/recovery/cubic/tests.rs#L167-L172). There's also `type=TODO` (which means you intend on implementing it but haven't done it yet) and `type=exception` (indicating that you don't plan on implementing it - usually a good idea to include `//= reason=This feature isn't needed because of X`).

Lastly, if you click on one of the highlighted requirements in the generated report, it displays a few buttons and clicking those will copy the comment to your clipboard.

PR #115, opened 2022-08-12
Maintain history of gh-pages

eagr:
At the moment, the updated workflow has a couple of problems:

* the gh-pages branch is always overwritten
* the job will fail when trying to commit but nothing is staged

Hopefully, this PR will fix those problems.

### DIFF
--- a/.github/actions/compliance/action.yml
+++ b/.github/actions/compliance/action.yml
@@ -61,8 +61,10 @@ runs:
         git config --local user.email "github-actions[bot]@users.noreply.github.com"
         git config --local user.name "github-actions[bot]"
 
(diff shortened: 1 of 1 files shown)


.github/actions/compliance/action.yml
```diff
@@ -72,7 +74,6 @@ runs:
     - name: Generate duvet report
       working-directory: ${{ inputs.dir }}
       run: |
-        ./ci/compliance/extract.sh
```
seanmonstar, line 75: Does this happen somewhere else?

eagr: Nah, this is the only place the scripts are executed.

seanmonstar (approved):

PR #116, opened 2022-08-13
Generate links to spec sections

eagr:
One last fix for the compliance report.

Just realized that the target argument to `duvet extract` is also used as the **reference/link** in the comments that are generated from the compliance report, which should be used as is so that the comment will be treated as an effective citation. I guess for ergonomics, we should probably pass the spec link to `duvet extract`, instead of the local path of the spec txt. Am I missing anything? @camshaft

But then the action will need to fetch the spec page on every run. We could commit the extracted specs instead, if you prefer otherwise. @seanmonstar

### DIFF
--- a/.github/actions/compliance/action.yml
+++ b/.github/actions/compliance/action.yml
@@ -62,9 +62,9 @@ runs:
         git config --local user.name "github-actions[bot]"
 
         if git ls-remote --exit-code --heads origin gh-pages; then
-          cp -r .github ci h3 target
+          cp -r .github ci h3 tests target
           git switch gh-pages
-          cp -r target/.github target/ci target/h3 .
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):

eagr:
Right now, links in the generated report are still somewhat broken. [Submitted some fix](https://github.com/awslabs/duvet/pull/68) for the problems, waiting for an additional approval. Also, we might want this feature called `implication` type from the newer version of duvet to generate a more accurate report.

So more work needs to be done to kinda finalize the report stuff. Just let me know when you decide to proceed with migrating to the newer version duvet. @seanmonstar

seanmonstar:
Whenever you want to migrate, it looks like it's a good call for the project! 🙏

PR #119, opened 2022-08-26
Feed compliance report generator

eagr:
Redoing #117 
* fixed broken anchors in spec links
* corrected some dislocated annotations
* applied a new annotation type [IMPLICATION](https://github.com/awslabs/duvet/blob/681f2b02178b2f9b98439ab4d9ff745ebbdf2375/src/annotation.rs#L142) for requirements whose related code is self-evident

### DIFF
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -28,7 +28,6 @@ where
     O: quic::OpenStreams<Bytes>,
 {
     //= https://www.rfc-editor.org/rfc/rfc9114#section-3.3
-    //= type=implication
     //# Clients SHOULD NOT open more than one HTTP/3 connection to a given IP
     //# address and UDP port, where the IP address and port might be derived
     //# from a URI, a selected alternative service ([ALTSVC]), a configured
@@ -79,7 +78,6 @@ where
         let headers = Header::request(method, uri, headers)?;
 
         //= https://www.rfc-editor.org/rfc/rfc9114#section-4.1
-        //= type=implication
         //# A
         //# client MUST send only a single request on a given stream.
         let mut stream = future::poll_fn(|cx| self.open.poll_open_bidi(cx))
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -193,12 +193,6 @@ where
 {
     pub async fn shutdown(&mut self, max_requests: usize) -> Result<(), Error> {
         self.inner.shutdown(max_requests).await
-
-        //= https://www.rfc-editor.org/rfc/rfc9114#section-5.2
-        //= type=TODO
-        //# An endpoint that completes a
-        //# graceful shutdown SHOULD use the H3_NO_ERROR error code when closing
-        //# the connection.
     }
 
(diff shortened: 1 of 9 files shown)


eagr:
progress so far:
image

seanmonstar:
Wonderful work!

eagr:
Ready for review! @seanmonstar 

There are many spec requirements left not annotated.
* some are not applicable, like IANA registration requirements
* some are kind of out of scope, like downgrading to a TCP-based protocol when fail to establish a QUIC connection
* others could be annotated as TODOs, which should be placed where the missing implementations should go. I don't have a clear idea about that so I guess I should probably leave them to someone who knows better.

Kinda wish there would be some `opt-out` type in duvet so we could have more intuitive stats, like in terms of how many applicable requirements left to do.

eagr:
Would you like to also bump the msrv in ci?

seanmonstar:
I bumped the MSRV just now, if you wanna rebase.

seanmonstar:
> some are not applicable, like IANA registration requirements

Do you have an example? I agree that likely we can skip these.

> some are kind of out of scope

Indeed, some are requirements that are at the application-level. Similar to how in the HTTP/2 spec, there's words about what to do if proxying HTTP/1 to HTTP/2, at the library level we can't know you're doing that. Should be fine to punt. Is there a decent way to do so?

Curious if @camshaft knows the best way we could signal that to duvet.

seanmonstar (approved):

camshaft:
In s2n-quic we've used `exception` for things that don't apply to the library (like IANA registration requirements). I think it captures the nature of it, especially since you can give a reason for why it's an exception.

eagr:
> Do you have an example? I agree that likely we can skip these.

https://www.rfc-editor.org/rfc/rfc9114#section-11.2.1

```
While this registry is separate from the "HTTP/2 Frame Type" registry
defined in [HTTP/2], it is preferable that the assignments parallel
each other where the code spaces overlap.  If an entry is present in
only one registry, every effort SHOULD be made to avoid assigning the
corresponding value to an unrelated operation.  Expert reviewers MAY
reject unrelated registrations that would conflict with the same
value in the corresponding registry.
```

> In s2n-quic we've used exception for things that don't apply to the library (like IANA registration requirements). I think it captures the nature of it, especially since you can give a reason for why it's an exception.

Thought about that too, but it didn't feel quite right. I guess I got too used to the concept of exception in programming. It kinda makes sense now. Thanks.

Say we proceed with using `type=exception`, where do you want to put the annotations in question? @seanmonstar @camshaft Since some of them have little connection with the code. They are basically just noises to the code base.

camshaft:
We have an `exceptions` directory with toml files

Example: https://github.com/aws/s2n-quic/blob/main/specs/exceptions/rfc8899/6.2.1.3.toml

eagr:
Oh, I see how this works, thanks! Will do it later in another pr.

Could we get this merged soon if it looks good? Kinda don't wanna linger over this for too long as there are many changes across the code base. Merge conflicts alert.

PR #122, opened 2022-09-01
WIP: Complete compliance report

eagr:

PR #123, opened 2022-09-02
ci: use published duvet

camshaft:
Now that duvet has been published to crates.io, it's better to install from there.

### DIFF
--- a/.github/actions/compliance/action.yml
+++ b/.github/actions/compliance/action.yml
@@ -16,12 +16,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/checkout@v3
-      with:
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):

eagr:
https://github.com/hyperium/h3/pull/122/commits/d018034d16352eef2927bfc7e1cdad3c40211f31

I have something hacked out that caches the compiled binary directly. Curious does the install action do that too? @camshaft

camshaft:
Yep! It checks the latest version and tries to retrieve it from [tool-cache](https://github.com/actions/toolkit/tree/master/packages/tool-cache). If it's not there then it builds from source and puts in in `tool-cache` with the built version.

PR #127, opened 2022-09-14
WIP: Complete the compliance report

eagr:
This PR aims to provide a more complete compliance report. Hopefully, this could help paint a more detailed picture of what is left to be done in h3.

### DIFF
--- a/ci/compliance/report.sh
+++ b/ci/compliance/report.sh
@@ -7,6 +7,7 @@ BLOB=${1:-master}
 duvet \
   report \
   --spec-pattern 'specs/**/*.toml' \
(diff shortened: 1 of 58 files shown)


seanmonstar (approved):
🤩

eagr:
Some requirements don’t apply to h3 directly, but

* the implementation of which may need some support from h3
* for some, it may be appropriate for h3 to guarantee the correctness

They all go into *todo*.

I kinda took the "better wrong than letting the requirement slip" approach, so some may actually belong to *exception*. Also, the reasons given by the exception annotations are rough. More work need to go in there..

PR #130, opened 2022-09-14
generate a sslkeylogfile

Ruben2424:
I changed the default tls settings in the example client to generate a `sslkeylogfile` if the `SSLKEYLOGFILE` environment variable is set.
This can help to understand the http/3 protocol or to debug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,12 +109,6 @@ As mentioned, the goal of this library is to be generic over a QUIC implementati
 This crate as well as the quic implementations are tested ([quinn](https://github.com/quinn-rs/quinn-interop), [s2n-quic](https://github.com/aws/s2n-quic/tree/main/scripts/interop)) for interoperability and performance in the [quic-interop-runner](https://github.com/marten-seemann/quic-interop-runner).
 You can see the results at (https://interop.seemann.io/).
 
-## Debugging
-The example [example client](https://github.com/hyperium/h3/examples/client.rs) can generate a `SSLKEYLOGFILE` to see the traffic unencrypted in tools like Wireshark.  
-To set this up just set the `SSLKEYLOGFILE` environment variable to a file path and follow this [tutorial](https://wiki.wireshark.org/TLS#using-the-pre-master-secret).
-Then use the example client with the `--keylogfile=true` option to enable this.
-
-
 ## License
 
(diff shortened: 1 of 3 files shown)


seanmonstar:
Is there any reason to not include this? Why doesn't it exist by default in rustls? Are there security concerns?

Ruben2424:
In Rustls this is disabled by default see: https://github.com/rustls/rustls/pull/868#issuecomment-988977923
Maybe this could be behind a command line argument, for example `--keylogfile=true`.
Or as path `--keylogfile=keylog/keylog.log`, but i don´t know if that is possible.

seanmonstar:
Ah yes, that's why I had remembered this before, I was pinged about it in that issue regarding reqwest 😅 . Since this is an "application" (examples compile as binaries), I think it's fine to provide the ability to turn it on... these two ideas come to my mind:

1. Just enable the option, and put a big code comment above "WARNING, we enable this for the example, you should think carefully about enabling in your own code."
2. Or, require a `--keylogfile` option before we turn it on, like you suggested.

Ruben2424:
> Ah yes, that's why I had remembered this before, I was pinged about it in that issue regarding reqwest sweat_smile . Since this is an "application" (examples compile as binaries), I think it's fine to provide the ability to turn it on... these two ideas come to my mind:
> 
>     1. Just enable the option, and put a big code comment above "WARNING, we enable this for the example, you should think carefully about enabling in your own code."
> 
>     2. Or, require a `--keylogfile` option before we turn it on, like you suggested.

I implemented both options.

README.md
```diff
@@ -109,6 +109,12 @@ As mentioned, the goal of this library is to be generic over a QUIC implementati
 This crate as well as the quic implementations are tested ([quinn](https://github.com/quinn-rs/quinn-interop), [s2n-quic](https://github.com/aws/s2n-quic/tree/main/scripts/interop)) for interoperability and performance in the [quic-interop-runner](https://github.com/marten-seemann/quic-interop-runner).
 You can see the results at (https://interop.seemann.io/).
 
+## Debugging
```
seanmonstar, line 112: Since this is only relevant to the examples, I'd put it in the `examples/README.md` instead.

seanmonstar (approved):

PR #131, opened 2022-09-15
document the duvet usage in h3

Ruben2424:
Hi, 
i have documented the duvet usage.
I hope that all i wrote is correct.

### DIFF
--- a/DUVET.md
+++ b/DUVET.md
@@ -1,0 +1,40 @@
+# Duvet
+This is a guide to explain how [duvet][] is used in h3.
+
+[duvet]: https://crates.io/crates/duvet
+
+## General
+Duvet tracks all keywords (MUST, MUST NOT, SHOULD, SHOULD NOT and MAY) from an RFC.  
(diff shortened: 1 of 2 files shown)


eagr:
duvet is still under active development and the interface may change soon, as you could see [here](https://github.com/awslabs/duvet/pull/63) in the duvet-specification.  Maybe we could keep this open and wait until the duvet 0.2 is out and link to the reference instead.

Also, it may be better to go into some Contribute section, along with a reminder of updating the annotations when behavior has changed.

seanmonstar:
@eagr it's been a couple months, does it still seem like something that is going to change soon, or should we just merge Ruben's work here?

eagr:
duvet has released a 0.2 but sadly the docs haven't landed, so we can't link to it yet. And the documentation pr seems a bit stagnant. Let's just do this.

Mind moving the docs into a CONTRIBUTING.md and refer to it from the README so that it can be more discoverable and less likely to be skipped by people? @Ruben2424

Ruben2424:
@eagr I renamed it.

eagr (approved):

PR #137, opened 2022-10-08
Improve compliance ci action

eagr:
* prep extraction script for multi-spec report
* use short hash for commit msgs
* clarify stuff a bit

Right now every push to master triggers a push to gh-pages even if the compliance status doesn't change, because the report generation isn't deterministic. Already fixed in duvet, (hopefully) the problem will auto resolve once the patches are released.


seanmonstar (approved):
Very cool!

PR #138, opened 2022-10-09
No more repetitive reports

eagr:
I could confirm that this patch fixes the problem on another fork.

image

This isn't ideal as source links in future reports may not always link to the right spot. But the latest report is alright. I guess this wouldn't be too much of a problem as we basically only care about the latest one.

### DIFF
--- a/.github/actions/compliance/action.yml
+++ b/.github/actions/compliance/action.yml
@@ -56,7 +56,7 @@ runs:
       shell: bash
       run: |
         ${{ inputs.extract-script }}
-        ${{ inputs.report-script }} $SHA
+        ${{ inputs.report-script }}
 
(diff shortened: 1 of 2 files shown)


camshaft:
If you only care about the latest report why not just do a force push with a single-commit history?

eagr:
We could, but what are the benefits of doing that? One of the reasons I want to do this is to quiet down the github-action bot. Right now every time we push to master, the bot would in turn push to gh-pages and puts up a notice. h3 isn't quite busy right now. If someday it gets updated more frequently, the bot could be spamming the org. dashboard.. That might be annoying to people in the org

seanmonstar (approved):

PR #139, opened 2022-10-10
Wrap up compliance report

eagr:
* refine wording of exception reasons
* move some exceptions into todo (maybe)

I guess we could start going through the todos and decide what to do next.


seanmonstar (approved):
<3

PR #145, opened 2022-11-12
Upgrade quinn

eagr:
Some breaking changes in Quinn@0.9.0:
* abstract rumtime support: need to include the runtime-tokio feature to use Tokio (if default-features is disabled)
* `Incoming` is replaced by methods in `Endpoint` and so the `Endpoint` constructors no longer return a tuple
* `NewConnection` is replaced by methods in `Connection`

Changes in h3-quinn:
* `SendStreamError` -> `SendError`
* `ReadError` -> `RecvError`


eagr:
This is tricky. 🤔 The tests are passing on my mac.

inflation:
On my Mac, the tests are failing too: 
```
failures:

---- tests::connection::control_close_send_error stdout ----
thread 'tests::connection::control_close_send_error' panicked at 'assertion failed: `Application { code: H3_STREAM_CREATION_ERROR, reason: Some("got two control streams"), level: ConnectionError }` does not match `Kind::Application {
reason: Some(reason), code: Code::H3_CLOSED_CRITICAL_STREAM, .. }
if *reason == *"control stream closed"`', h3/src/tests/connection.rs:335:9

---- tests::connection::graceful_shutdown_closes_when_idle stdout ----
thread 'tests::connection::graceful_shutdown_closes_when_idle' panicked at 'assertion failed: `Err(Elapsed(()))` does not match `Ok(())`', h3/src/tests/connection.rs:691:16

---- tests::request::get stdout ----
thread 'tests::request::get' panicked at 'accept: h3::Error { timeout: true }', h3/src/tests/request.rs:57:74

---- tests::request::header_too_big_server_error stdout ----
thread 'tests::request::header_too_big_server_error' panicked at 'accept: h3::Error { timeout: true }', h3/src/tests/request.rs:688:74


failures:
    tests::connection::control_close_send_error
    tests::connection::graceful_shutdown_closes_when_idle
    tests::request::get
    tests::request::header_too_big_server_error
```

This is on AArch64 macOS Ventura 13.0.1.

h3-quinn/src/lib.rs
```diff
@@ -1,223 +1,149 @@
...
-        let (send, recv) = match ready!(self.incoming_bi.next().poll_unpin(cx)) {
-            Some(x) => x?,
-            None => return Poll::Ready(Ok(None)),
-        };
-        Poll::Ready(Ok(Some(Self::BidiStream {
-            send: Self::SendStream::new(send),
-            recv: Self::RecvStream::new(recv),
-        })))
+        Poll::Ready(
+            match ready!(Box::pin(self.conn.accept_bi()).poll_unpin(cx)) {
```
inflation, line 45: How about pining it on the stack:

```rust
let mut future = self.conn.accept_bi();
let future = unsafe { Pin::new_unchecked(&mut future);
```

And poll the future directly:

```rust
future.poll(cx)
```

This is essentially what `pin_mut!` does.

h3-quinn/src/lib.rs
```diff
@@ -1,223 +1,149 @@
...
+                Ok((send, recv)) => Ok(Some(Self::BidiStream::new(
+                    Self::SendStream::new(send),
+                    Self::RecvStream::new(recv),
+                ))),
```
inflation, line 49: Note that there is a situation that returns `Ready(Ok(None))`, and checked inside `IncomingStream::poll_next` before, which is when the error is `ConnectionError::LocallyClosed`. 

We may choose to handle that in `h3` directly though.

h3-quinn/src/lib.rs
```diff
@@ -257,189 +181,95 @@ where
...
-                            err.into_inner()
-                                .expect("write stream returned an empty error")
-                                .downcast_ref::<WriteError>()
-                                .expect(
-                                    "write stream returned an error which type is not WriteError",
-                                )
-                                .clone(),
-                        )));
+                let mut write_fut = Box::pin(self.stream.write(data.chunk()));
+                match ready!(write_fut.poll_unpin(cx)) {
```
inflation, line 245: You could use a block here: 
```rust
match ready!({
    let mut write_fut = self.stream.write(data.chunk());
    let write_fut = unsafe { Pin::new_unchecked(&mut write_fut)) };
    write_fut.poll(cx)
})
```

So, you don't need to explicitly drop that future.

h3-quinn/src/lib.rs
inflation: Adding that fixes everything else except `tests::connection::control_close_send_error`.

eagr:
> On my Mac, the tests are failing too:
> 
> ```
> failures:
> 
> ---- tests::connection::control_close_send_error stdout ----
> thread 'tests::connection::control_close_send_error' panicked at 'assertion failed: `Application { code: H3_STREAM_CREATION_ERROR, reason: Some("got two control streams"), level: ConnectionError }` does not match `Kind::Application {
> reason: Some(reason), code: Code::H3_CLOSED_CRITICAL_STREAM, .. }
> if *reason == *"control stream closed"`', h3/src/tests/connection.rs:335:9
> 
> ---- tests::connection::graceful_shutdown_closes_when_idle stdout ----
> thread 'tests::connection::graceful_shutdown_closes_when_idle' panicked at 'assertion failed: `Err(Elapsed(()))` does not match `Ok(())`', h3/src/tests/connection.rs:691:16
> 
> ---- tests::request::get stdout ----
> thread 'tests::request::get' panicked at 'accept: h3::Error { timeout: true }', h3/src/tests/request.rs:57:74
> 
> ---- tests::request::header_too_big_server_error stdout ----
> thread 'tests::request::header_too_big_server_error' panicked at 'accept: h3::Error { timeout: true }', h3/src/tests/request.rs:688:74
> 
> 
> failures:
>     tests::connection::control_close_send_error
>     tests::connection::graceful_shutdown_closes_when_idle
>     tests::request::get
>     tests::request::header_too_big_server_error
> ```
> 
> This is on AArch64 macOS Ventura 13.0.1.

Yeah, on my machine, the control_close_send_error test is failing, but that's known before the upgrade.

h3-quinn/src/lib.rs
eagr: nice catch!

eagr:
The tests passed on macOS but failed on Ubuntu due to connection timeout (at least that's what the errors said). We could make the tests pass by calling `poll_finish()`. But I'm not sure if it's the real problem.

Having revisited this a couple of times, I still couldn't figure out why the connections timed out. And It's tricky for me to debug this since I don't have a Linux machine. Any ideas? @seanmonstar @inflation

Ruben2424:
I have a Linux machine. I can try to find the Issue.

Ruben2424:
Ok i can reproduce the error. But it appears not always. I will debug to find the exact issue but i can not say how much time I need to find it.

h3-quinn/src/lib.rs
seanmonstar: I think this code causes a problem, since it's creating a future (calling an async fn) on poll, polling it once, and if it wasn't ready, the future is dropped since it's in the function stack. To translate correctly to polling, the future would need to be store in self somehow.

Ruben2424: > I think this code causes a problem, since it's creating a future (calling an async fn) on poll, polling it once, and if it wasn't ready, the future is dropped since it's in the function stack. To translate correctly to polling, the future would need to be store in self somehow.
> 
> This problem is in each of the poll functions.

This is probably the reason why the connections timed out. Every time I tested this and the future was ready it worked. If the future was not ready it timed out.

inflation:
Can confirm this fixed most tests, but the `server-client` example tests still timed out after the changes. 

PS. `cargo nextest` can archive tests and run them somewhere else. This really helps in this situation. I could cross-compile the tests to Linux and run them in docker.

h3-quinn/src/lib.rs
eagr: @seanmonstar I see. It totally makes sense. I feel I may need further guidance for this. I don't see a clear path for storing the futures in self without changing the traits as they require lifetime annotations now.

Curious, why did the tests pass on macOS?

eagr:
Yeah, coz I only fixed the tests in /h3. Making the tests pass like this doesn't address the real problem anyway. Will check that out. Thanks! @inflation

h3-quinn/src/lib.rs
inflation: So, we need to somehow “park” the futures. It looks like what an async executor need to do, so I wonder if `tokio` would provide some tools for that.

Ruben2424: > @seanmonstar I see. It totally makes sense. I feel I may need further guidance for this. I don't see a clear path for storing the futures in self without changing the traits as they require lifetime annotations now.
> 
> Curious, why did the tests pass on macOS?

what would speak against changing the traits? The crate is not released yet, right?

Ruben2424:
Oops, i wanted to merge the changes in a lokal branch and accidentally committed this change her. I´m Sorry.

Ruben2424:
How can i Undo this?

inflation:
Reset and force push souls be fine if you're working alone.

Ralith:
Please feel free to reach out to the Quinn project if you're having difficulty with the API, also.

PR #150, opened 2022-12-05
Separate & simplify Settings impl.

eagr:
* get rid of `SettingId`
* separate Settings frame implementation
* add `ENABLE_WEBTRANSPORT` setting type

After having moved `is_reserved(id)` and `is_supported(id)` out, I noticed that the `SettingId` abstraction was doing very little. I felt it might not worth the overhead. And using it alongside `Settings` kinda exposes the internal representation of the settings. I think removing it could make the code simpler and easier to change.


Ruben2424:
I would keep the SettingsId struct with or without the `is_reserved(id)` and `is_supported(id)`. 

Then the Ids are abstracted away from the Settings struct. `SettingsID` represents one Setting and the `Settings` holds multiple Settings. 

And instead of moving `is_reserved(id)` and `is_supported(id)` to `Settings` we could call them inside `SettingsID` when the struct is created.

eagr:
I guess I just don't see a good reason of adding an extra layer over the ids. What do we really miss without it?

eagr:
Let's look at the code. Why do `is_supported(&self, id)` when we could just do `is_supported(id)`, that is, we don't really need the wrapping for that. And `encode` and `decode` are basically just redirections here. Also, we need to wrap and unwrap ids for each setting (when we could choose to avoid them).

We need a good reason to justify those. Right now, I just don't see it.

Ruben2424:
I'm sorry but I don't see it too. In my opinion it is better to separate these things. 
I would leave the `is_frobidden` in `SettingsId` and call it when constructing. Maybe something like that.

```rs
impl SettingId {
    pub fn from(id :u64) -> Result<Self, SettingsError> {
        if !is_forbidden(id) {
            Ok(Self(id))
        }else {
            Err(SettingsError::InvalidSettingId(id))
        }
    }
}
```

For me that is more readable because there are 2 smaller structs instead of one bigger. And it also supports change better because the `SettingsID` don't care about `Settings` and `Settings` only cares about the public API in `SettingsID`. Maybe we need another opinion here.

eagr:
We could achieve separation of concerns with functions too. Try to think of `Settings` as a namespace for grouping related functions instead. Grouping those functions into `Settings` makes sense to me.

Alright, I was doing some experimentation on wt and saw stuff that just felt odd.

```rs
fn is_forbidden(&self) -> bool {
    matches!(
        self,
        SettingId(0x00) | SettingId(0x02) | SettingId(0x03) | SettingId(0x04) | SettingId(0x05)
    )
}
```

The function could simply accept the a setting id as it is (an integer) right? And it would probably make the function more useful as integer is a more universal interface, like

```rs
fn is_forbidden(id: u64) -> bool {
    matches!(id, 0 | 2 | 3 | 4 | 5)
}
```

And once I made `is_supported` and `is_reserved` functions instead of methods, I realized that there isn't much in `SettingId`. Getting rid of it felt right to me, as it seemed to just complicate stuff for no good reasons.

I didn't intend this to be dreadful. I just felt it could simplify stuff a bit, might be nice to also upstream these changes. It's alright if this doesn't go into h3. Not a big deal.

eagr:

Issue #71, opened 2021-12-24
Add WebTransport support


Pascualex:
It would be useful to support WebTransport now that it is [getting released in Chrome 97](https://blog.chromium.org/2021/11/chrome-97-webtransport-new-array-static.html).

After looking into it, some additional work seems to be required to enable this support.

The [Session Establishment section of the draft](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-3) defines some required settings for the settings frame. I've been experimenting a bit but I'm still stuck with a `net::ERR_METHOD_NOT_SUPPORTED` and an unsuccessful handshake.

I'm not sure if after achieving a successful handshake there will be additional work required to support this protocol.

I'm using [aioquic](https://github.com/aiortc/aioquic) as a reference, since that is the library that Chrome is using to support their [example WebTransport server](https://github.com/GoogleChrome/samples/blob/7380e933887c1c3f03dac62eeb042b6d6e7d297d/webtransport/webtransport_server.py).

stammw:
Hi @Pascualex,

I wanted to come up with an exhaustive list of what's missing in this crate to support WebTransport. But with the holidays I actually don't have much time for this. So here are some points I found, which will make the WebTransport implementation a non-trivial work:

- Custom Settings, namely for `SETTINGS_ENABLE_WEBTRANSPORT`.
- CONNECT method has to be implemented.
- Bidirectionnal stream from server
- Custom stream types
- Custom frame types
- Custom Error codes
- HTTP Datagrams

Neqo seems to implement directly web-transport into their http3 crate, but as h3 goal is to be highly reusable for a variety of use cases, we might want to implement some of these features in a separate crate and offer an extensibility methods within `h3`. (I assume the main goal of Neqo is to offer the http3 in Firefox only).

We could also implement WebTransport with feature-gated code within h3. But the extensibility seems a rather future-proof approach for other public or vendor protocol extensions. 

Nevertheless, I think we'll concentrate on this starting at v0.2 the soonest. If your keen to help us you are more than welcome! :)

Pascualex:
Hi @stammw,

Thanks for looking into it. I agree that this should not be part of the core of H3 and that it makes sense to delay this until the core is released.

To be honest with you what I know of these protocols is mostly from a user perspective, but if there is something I can help with I'll gladly do it.

zhenzhu478:
> Hi @Pascualex,
> 
[...]

PR #151, opened 2022-12-05
HTTP/3 connection params

eagr:
Dedup client/server builders. And it's likely there will be more shared parameters in the future.

### DIFF
--- a/h3/src/params.rs
+++ b/h3/src/params.rs
@@ -1,0 +1,45 @@
+//! HTTP/3 connection parameters
+
+/// HTTP/3 connection parameters builder
+pub struct Params {
+    pub(crate) enable_webtransport: bool,
+    pub(crate) grease: bool,
+    pub(crate) max_field_section_size: u64,
+}
+
+impl Default for Params {
+    fn default() -> Self {
+        Self {
+            enable_webtransport: false,
+            grease: true,
+            max_field_section_size: Self::DEFAULT_MAX_FIELD_SECTION_SIZE,
+        }
+    }
+}
+
+impl Params {
+    /// Default max header size
+    pub const DEFAULT_MAX_FIELD_SECTION_SIZE: u64 = (1 << 62) - 1;
+
+    /// Set whether WebTransport is supported
+    pub fn enable_webtransport(mut self, val: bool) -> Self {
+        self.enable_webtransport = val;
+        self
+    }
+
+    /// Set wether to send GREASE
+    pub fn grease(mut self, val: bool) -> Self {
+        self.grease = val;
+        self
+    }
+
+    /// Set the maximum header size the endpoint is willing to accept
+    ///
+    /// See [header size constraints] section of the specification for details.
+    ///
+    /// [header size constraints]: https://www.rfc-editor.org/rfc/rfc9114.html#name-header-size-constraints
+    pub fn max_field_section_size(mut self, val: u64) -> Self {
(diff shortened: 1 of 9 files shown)


eagr:
Should we keep the client/server builders? Also, any suggestions on the name of the parameter builder?

Ruben2424:
> Should we keep the client/server builders?

I would keep the client/server builders because there will possibly be parameters which only belong to server or client.

>  Also, any suggestions on the name of the parameter builder?

Maybe just `ParameterBuilder`? But i'm not good with names.

h3/src/params.rs
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+}
+
+impl Params {
+    /// Default max header size
+    pub const DEFAULT_MAX_FIELD_SECTION_SIZE: u64 = (1 << 62) - 1;
+
+    /// Set whether WebTransport is supported
+    pub fn enable_webtransport(mut self, val: bool) -> Self {
```
Ruben2424, line 25: It might be better to put the WebTransport stuff behind a Feature Flag. What do you think?

eagr: Yeah, I thought about that too. But what if it's just for params and settings?

Ruben2424: I would put WebTransport completely behind a feature flag or maybe even in a separate crate. Iirc there was some discussion about this in #71.

eagr: I understand the benefits of putting stuff behind a feature gate sometimes (smaller binary, faster compilation, etc). I totally agree we should feature-gate or separate major implementations of WT.

But configs and settings are lightweight and hard to separate. The benefit of putting those behind a feature gate seems limited. I'm 100% sure about this though. What do you think? @seanmonstar

seanmonstar: I have not had the time to read the WebTransport spec, so not sure what more is needed, but: does adding this at least allow for WebTransport to be _proxied_? If it's needed for a proxy to simply pass the data frames along, then seems fair to include it as part of the crate, perhaps?

eagr: Yeah, I think so. The setting would be required to be sent by both sides when establishing connection to create a WT session.

eagr:
The API still feels a bit weird. There should probably be one builder instead of two: `Config` and `Builder`.

```rs
h3::config::Config {}

h3::client::Builder { config }
h3::server::Builder { config }
```

So, it should either be something that looks like

```rs
// somehow dedupped
h3::config::ClientConfig {}
h3::config::ServerConfig {}

// no more Builder
h3::client::new(conn, client_config)
h3::server::new(conn, server_config)
```

or

```rs
// somehow dedupped
h3::Endpoint::client().build(conn)
h3::Endpoint::server().build(conn)
```

Ruben2424:
I like this option better.
```rs
// somehow dedupped
h3::config::ClientConfig {}
h3::config::ServerConfig {}

// no more Builder
h3::client::new(conn, client_config)
h3::server::new(conn, server_config)
```

eagr:
Yeah. Using a config struct has a potential advantage of easier mapping to settings. We could potentially have something like `impl From<Config> for Settings`.

eagr:
Changed user-facing config structs to be compositions: 

```rs
ClientConfig { conn: ConnectionConfig, ... }
ServerConfig { conn: ConnectionConfig, ... }
```

this way we could could easily provide some config for the shared `ConnectionInner`.

Issue #51, opened 2021-09-20
Error when running server example


SergeiMinaev:
```
$ cargo run --example server selfsigned 
Finished dev [unoptimized + debuginfo] target(s) in 0.19s
Running target/debug/examples/server selfsigned
```

I'm trying to connect via curl:
```
$ curl --http3 https://localhost:4433
```
After that, an error occurs:
```
$ cargo run --example server selfsigned 
    Finished dev [unoptimized + debuginfo] target(s) in 0.21s
     Running `target/debug/examples/server selfsigned`
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: h3::Error { kind: Transport, cause: ApplicationClosed(ApplicationClose { error_code: 0, reason: b"" }) }', examples/server.rs:124:80
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

If I try to connnect via `cargo run --example client` there's another error:
```
$ cargo run --example client https://localhost:4433
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/examples/client 'https://localhost:4433'`
DNS Lookup for https://localhost:4433/: 127.0.0.1:4433
Error: ConnectionClosed(ConnectionClose { error_code: Code::crypto(78), frame_type: None, reason: b"peer doesn't support any known protocol" })

```

seanmonstar:
Ah, looks like the server example has an old set of ALPN protocols set. When I locally updated that to `h3-29`, it worked.

SergeiMinaev:
I've changed `server_config.protocols(&[b"h3-27"]);` to `server_config.protocols(&[b"h3-29"]);`.
After that, the client example returns another error:
```
$ cargo run --example client https://localhost:4433
    Finished dev [unoptimized + debuginfo] target(s) in 0.26s
     Running `target/debug/examples/client 'https://localhost:4433'`
DNS Lookup for https://localhost:4433/: 127.0.0.1:4433
Error: TransportError(Error { code: Code::crypto(2a), frame: None, reason: "invalid certificate: UnknownIssuer" }
```

But when I connecting with curl, tokio is still panicking:
```
$ cargo run --example server selfsigned 
   Compiling examples v0.0.0 (/media/Program/rust/h3/examples)
    Finished dev [unoptimized + debuginfo] target(s) in 7.94s
     Running `target/debug/examples/server selfsigned
[...]

PR #152, opened 2022-12-07
Clean up client/server examples

eagr:
Examples should be as straight-forward as possible and focused on demonstrating the API usage. Trying to remove the non-essential clouding parts from the examples to make them easier to understand.


eagr:
Tempted to remove the `--insecure` flag from the client example, as it doesn't really demo how h3 should be used. Is it there for a particular reason?

inflation:
One more thing: the integration test leaks if it panics when testing, so adding an RAII struct to clean up the server would be great.

seanmonstar:
> Tempted to remove the --insecure flag from the client example, as it doesn't really demo how h3 should be used. Is it there for a particular reason?

I don't have a strong opinion, only that the examples currently provide two things: 1) how to _use_ `h3`, and 2) an easy way to check interop with a different implementation.

So, it'd be good if it's still very easy for a newcomer to check "does this actually work with cloudflare or youtube".

eagr:
> I don't have a strong opinion. I only note that the examples currently provide two things: 1) how to use h3, and 2) an easy way to check interop with a different implementation.

Noted.

I guess the flag `--insecure` was added to address this issue #51. If we use a CA certificate instead of a self-signed one, that wouldn't be a problem right?

eagr:
To sum it up a bit, hopefully by uploading a CA cert, we could get rid of some non-essential logic from the examples

* `--insecure`
* dynamically generating self-signed cert when not provided

to make them more straight-forward and easier to learn from.

examples/client.rs
```diff
@@ -31,76 +33,85 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            .with_custom_certificate_verifier(Arc::new(YesVerifier))
-            .with_no_client_auth()
+        }
+        Err(e) => {
+            error!("couldn't load any default trust roots: {}", e);
+        }
     };
+
+    // load certificate of CA who issues the server certificate
+    if let Err(e) = roots.add(&rustls::Certificate(std::fs::read(opt.ca)?)) {
```
Ruben2424, line 78: Good job the examples are simpler now.

Do we need a warning here that we do that for the example, but to think twice before doing that in production code?
Also iirc this may enable Man-in-the-middle attack when using the example client.

eagr: done

eagr:
Tempted to remove tests against nightly from CI. They probably shouldn't be run on a per-PR basis.

PR #156, opened 2022-12-12
CI overhaul

### LINKED ISSUE #154: Remove nightly tests from CI?
Compiling tests against nightly Rust sometimes fails. We could choose to ignore it, but if it fails before other tests finish, it would cancel other tests.

If there's no particular reasons that we want to build/test against nightly, it seems better to remove them, at least, not let them run on a per-pr basis.

eagr:
closes #154 

* remove nightly tests to unblock stuff
* nuke `build` job
* cargo check MSRV
* pin to nightly-2022-12-01
* cache builds (esp. cargo-fuzz) to speed things up

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,9 +9,9 @@ env:
   RUST_BACKTRACE: 1
   toolchain_style: stable
   toolchain_msrv: 1.56
-  toolchain_doc: nightly-2022-11-03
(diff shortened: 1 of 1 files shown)


eagr:
The CI jobs finish much more quickly now.

seanmonstar (approved):

seanmonstar:
Squash and merge when you like it!

PR #157, opened 2022-12-13
Simplify client/server examples

eagr:
* Get rid of non-essential logic that may make the examples harder to follow by uploading server and CA certificates (for example only). Hopefully, this could also bring the examples closer to how stuff should be done in practice.
* Make sure the integration test exits on panics by cleaning up child process.


examples/client.rs
```diff
@@ -31,76 +33,86 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            .with_no_client_auth()
+        }
+        Err(e) => {
+            error!("couldn't load any default trust roots: {}", e);
+        }
     };
+
+    // load certificate of CA who issues the server certificate
+    // NOTE that this should be used for dev only
+    if let Err(e) = roots.add(&rustls::Certificate(std::fs::read(opt.ca)?)) {
```
Ruben2424, line 79: Looks really good.
I\`m not sure. But when using the client to connect to other services, would trusting this Certificate enable some sort of MITM? 
Maybe it would be better to make the default of `ca` something that doesn\`t load a Certificate. This could also make the Client more trustworthy. What do you think?

Ruben2424:
just curious, did no ci run?

examples/client.rs
eagr: I happened to apply some mitm trick a while ago, to trick some web frontend on my device into displaying tampered results (prank on my friend). If this is what you're talking about, in order to pull it off, the attacker will at least need
* to be somewhat in control of the traffic so they can intercept the messages
* to somehow plant their CA certificate (not this one) in the victim

eagr:
emm.. It's [triggered](https://github.com/hyperium/h3/actions/runs/3687754081), but is not associated to this pr

Ruben2424 (approved):

PR #158, opened 2022-12-14
Replace endpoint builders with config builders

eagr:
Introducing some config builders to replace current endpoint builders.

```rs
// should be passed to ConnectionInner
ConnectionConfig {
    // configs for the shared connection implementation
}

ClientConfig {
    conn: ConnectionConfig,
    // client-specific configs...
}

ServerConfig {
    conn: ConnectionConfig,
    // server-specific configs...
}
```

Potential benefits:
* config structs feel more self-contained and natural to pass around
* opens up opportunities to do conversions like `impl From<ConnectionConfig> for Settings`
* dedup current builder code

TODO
* Typing of `client::new()` and `server::new()` should be more generic, with `Buf` instead of `Bytes`. But there seems to some usability issue with `Buf`: the typing can't be inferred, have to use explicit typing on `new()`. It feels like the issues comes from h3-quinn. I'm not sure.
* A handy API for using default config, like `client::new_default(conn)` (probably not..).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,24 @@
 # h3
 
-An async HTTP/3 implementation.
-
 [![License: MIT](https://img.shields.io/badge/License-MIT-blue.svg)](LICENSE)
 [![CI](https://github.com/hyperium/h3/workflows/CI/badge.svg)](https://github.com/hyperium/h3/actions?query=workflow%3ACI)
 [![Discord chat](https://img.shields.io/discord/500028886025895936.svg?logo=discord)](https://discord.gg/q5mVhMD)
 
-This crate provides an [HTTP/3][spec] implementation that is generic over a provided QUIC transport. This allows the project to focus on just HTTP/3, while letting users pick their QUIC implementation based on their specific needs. It includes client and server APIs. Check the original [design][] for more details.
+An async HTTP/3 implementation.
+
+This crate provides an [HTTP/3][spec] implementation that is generic over a provided QUIC transport. This allows the project to focus on just HTTP/3, while letting users pick their QUIC implementation based on their specific needs. Check the original [design][] for more details.
+
+Features
+
+* async API
+* **[`h3`](./h3)** HTTP/3 client and server implementation
+* **[`quic`](./h3/src/quic.rs)** contains traits for abstracting QUIC implementations
+* **[`h3-quinn`](./h3-quinn)** integration support for Quinn
+
+As mentioned, the goal of this library is to be generic over a QUIC implementation. To that effect, integrations with QUIC libraries exist:
+
+- [`h3-quinn`](./h3-quinn/)
+- [`s2n-quic-h3`](https://github.com/aws/s2n-quic/tree/main/quic/s2n-quic-h3)
 
 [spec]: https://www.rfc-editor.org/rfc/rfc9114
 [design]: design/PROPOSAL.md
@@ -19,6 +31,13 @@ The eventual goal is to use `h3` as an internal dependency of [hyper][].
 
 [hyper]: https://hyper.rs
 
+## Getting Started
+
+The [examples](./examples) directory can help get started in two ways:
+
+- There are ready-to-use `client` and `server` binaries to interact with _other_ HTTP/3 peers. Check the README in that directory.
+- The source code of those examples can help teach how to use `h3` as either a client or a server.
+
 ### Duvet
 This create uses the [duvet crate][] to check compliance of the [spec][].  
 The generated [report][] displays the current status of the requirements of the spec.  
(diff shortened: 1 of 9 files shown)


eagr:
Hmm.. There seems to be a similar issue in beta Rust compiler.

Ruben2424:
> Hmm.. There seems to be a similar issue in beta Rust compiler.
It looks like the same issue.
I had a closer look at it and found out that this line (and the same line in other tests) causes the issue.
https://github.com/hyperium/h3/blob/885144bfcb0414bcb86b63cec9c65a611c7e56ee/h3/src/tests/connection.rs#L586

But I coudn\`t create a minimized example. Maybe it is the complex type of `stream` or something else.

eagr:
Maybe just try to report this with whatever info you have and see what the compiler team would ask for.

Ruben2424:
Is this ready for review?

eagr:
Yep, please do. Btw, happy new year!

h3/src/server.rs
```diff
@@ -118,13 +138,25 @@ where
...
+    /// Create a new HTTP/3 server connection
     ///
-    /// Use a custom [`Builder`] with [`builder()`] to create a connection
-    /// with different settings.
-    /// Provide a Connection which implements [`quic::Connection`].
-    pub async fn new(conn: C) -> Result<Self, Error> {
-        builder().build(conn).await
+    /// Provide a connection endpoint which implements [`quic::Connection`] and
+    /// desired configuration for the server.
+    pub async fn new(conn: C, config: ServerConfig) -> Result<Self, Error> {
```
Ruben2424, line 145: There is this `new` method and the `new` function in line 97. 
The `new` function just calls this method.

This `new` method is also public so there are two ways possible to create a server `Connection`
I would remove this `new` method and put the code also in the new function like it is in the client.

Ruben2424 (commented):
I like these changes. And a Happy New Year to you too!

h3/src/server.rs
eagr: I guess there is no harm of keeping this right?

Ruben2424:
Oh sorry I forgot to mention this in the review, but I think the examples in the readme should also be updated, right?

h3/src/server.rs
Ruben2424: Yes probably not

README.md
```diff
@@ -30,105 +48,6 @@ Get more information about this tool in the [contributing][] document.
...
-* QUIC transport abstraction via traits in the [`quic`](./h3/src/quic.rs) module
-* Supported QUIC implementations to date are
-  [Quinn](https://github.com/quinn-rs/quinn) ([h3-quinn](./h3-quinn/))
-  and [s2n-quic](https://github.com/aws/s2n-quic)
-  ([s2n-quic-h3](https://github.com/aws/s2n-quic/tree/main/quic/s2n-quic-h3))
-
-## Overview
-
-* **h3** HTTP/3 implementation
-* **h3-quinn** QUIC transport implementation based on [Quinn](https://github.com/quinn-rs/quinn/)
```
Ruben2424, line 46: Why did you remove the Overview?

README.md
```diff
@@ -30,105 +48,6 @@ Get more information about this tool in the [contributing][] document.
...
-
-## Getting Started
-
-The [examples](./examples) directory can help get started in two ways:
-
-- There are ready-to-use `client` and `server` binaries to interact with _other_ HTTP/3 peers. Check the README in that directory.
-- The source code of those examples can help teach how to use `h3` as either a client or a server.
-
-### Server
```
Ruben2424, line 55: I\`m not sure if we should leave the small examples in the readme.

eagr: The code to be pasted here would be snippets from the examples anyway. I assume a pointer to the examples should be enough to get people started right?

README.md
eagr: It's kinda merged into the intro part. Is there anything missing from the intro part? I did the cleanup coz it seems like people just jammed stuff into the README and it wasn't really maintained. The structure was kind messy. I structured it a little: overview, status, getting started, interop info

Ruben2424: This is the file structure, folder `h3` is the actual http/3 implementation. An in `h3-quinn` there is the quinn stuff.

I think it\'s good because it gives a quick overview of how the project is structured.

README.md
Ruben2424: I\`m absolutely not sure about this one. 
On the one hand this does\`t get tested or compiled so when api changes we have to remember to update this and the information provided from this is also in the examples and the docs.
On the other hand this gives a quick starting point when someone comes here first time without the need to find the right places in the examples or the docs. 
I would prefer to keep the snippets here bit I\`m OK with removing them as well.

Ruben2424: > That is really not critical. I'm actually more concerned about the API changes. User having to do `h3::client::new(conn, Default::default())` (the default part) kinda bugs me a little. Do you have any input on that?

Hmm good question, it is possible to call `h3::client::new(conn, ClientConfig::default())`. Another option is to make a `build` method for the Config structs which calls the `new` function, but this brings the older builder like style back.

All in all I do not have much experience in api design so I am not sure what is best here.

eagr:

PR #159, opened 2022-12-14
Enable extension negotiation

eagr:
This is an attempt to remove some roadblock to future extension support.

Both WebSocket/WebTransport over HTTP/3 rely on CONNECT requests bearing a `:protocol` pseudo-header to create new sessions. However, `:protocol` is deemed invalid in h3.

I'm assuming people would use `http::Extensions` to hold the value of the `:protocol` header. Is there a better option in `http` for this?

### DIFF
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -156,9 +156,10 @@ where
             method,
             uri,
             headers,
+            extensions,
             ..
         } = parts;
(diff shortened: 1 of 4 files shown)


seanmonstar:
Using an Extension is what `h2` has done.

eagr:

PR #173, opened 2023-03-10
Consolidate quic trait redundancy

FlorianUekermann:
The OpenStreams and SendStream quic traits overlap significantly. This has two effects if both are implemented on the same type:
1. Unnecessary repetition of identical implementations
2. Method ambiguity if both traits are imported, which makes calling redundant methods very awkward (need to specify trait, even though the methods actually do the same thing).

A convenient side effect is that this allows using different types for open errors and accept errors. (Most importantly closed quic connections typically return `Err(_)` on the `poll_open` methods, while returning `Ok(None)` on the `poll_accept` methods) and therefore include fewer variants.


Ruben2424 (approved):
Thanks for the PR! 
I really like this change.
Since this is changing the traits I would like to get a +1 from a second collaborator before merging.

Issue #175, opened 2023-03-12
Wrong length encoding in simple frames (CANCEL_PUSH, GOAWAY, MAX_PUSH_ID)


FlorianUekermann:
https://github.com/hyperium/h3/blob/6a2be4a21a6b9f1459bd023e83cfa7f2ee7a5b59/h3/src/proto/frame.rs#L327-L331

L329 is wrong, right? It should be `buf.write_var(id.size())`.

I need to touch that anyway to fix #174 , but wanted to confirm first.

PR #176, opened 2023-03-12
fix client shutdown and push id / stream id confusion

### LINKED ISSUE #174: push ids are not stream ids
The stream id type is used to represent and validate push ids. On GOAWAY frames from the client, `StreamId::is_push()` is used to validate the push id. Based on my understanding of the RFC I believe this is incorrect (they are not related) and it leads to errors on client initiated shutdowns:
```
h3::Error { code: H3_ID_ERROR, reason: "non-push StreamId in a GoAway frame: client bidirectional stream 0" }
```
I can send a PR with the following changes:
- add a PushId type
- Either:
  - differentiate Frame::ServerGoaway(StreamId) and Frame::ClientGoaway(PushId)
  - introduce `GoawayId` with stream and push variants for Frame::Goaway(GoawayId)
  - use Frame::Goaway(u64) and convert to respective type as soon as possible after unpacking

FlorianUekermann:
Fixes #174 and #175

Note: On top of this change ids could use a bit of a clean-up. They aren't used consistently and differ in how well initialization and conversion are protected from creating invalid values etc. But that is beyond the scope of this PR, I would like to address that in a follow-up.

### DIFF
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -16,8 +16,10 @@ use crate::{
     connection::{self, ConnectionInner, ConnectionState, SharedStateRef},
     error::{Code, Error, ErrorLevel},
     frame::FrameStream,
-    proto::{frame::Frame, headers::Header, varint::VarInt},
-    qpack, quic, stream,
+    proto::{frame::Frame, headers::Header, push::PushId, varint::VarInt},
+    qpack,
+    quic::{self, StreamId},
+    stream,
 };
 
 /// Start building a new HTTP/3 client
@@ -147,7 +149,7 @@ where
             (state.peer_max_field_section_size, state.closing)
         };
 
-        if closing.is_some() {
+        if closing {
             return Err(Error::closing());
         }
 
@@ -348,6 +350,10 @@ where
     B: Buf,
 {
     inner: ConnectionInner<C, B>,
+    // Has a GOAWAY frame been sent? If so, this PushId is the last we are willing to accept.
+    sent_closing: Option<PushId>,
+    // Has a GOAWAY frame been received? If so, this is StreamId the last the remote will accept.
+    recv_closing: Option<StreamId>,
 }
 
 impl<C, B> Connection<C, B>
@@ -355,9 +361,10 @@ where
     C: quic::Connection<B>,
     B: Buf,
 {
-    /// Itiniate a graceful shutdown, accepting `max_request` potentially in-flight server push
-    pub async fn shutdown(&mut self, max_requests: usize) -> Result<(), Error> {
-        self.inner.shutdown(max_requests).await
+    /// Initiate a graceful shutdown, accepting `max_push` potentially in-flight server pushes
+    pub async fn shutdown(&mut self, _max_push: usize) -> Result<(), Error> {
+        // TODO: Calculate remaining pushes once server push is implemented.
+        self.inner.shutdown(&mut self.sent_closing, PushId(0)).await
     }
 
     /// Wait until the connection is closed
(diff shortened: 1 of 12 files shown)


h3/src/connection.rs
```diff
@@ -460,8 +433,43 @@ where
...
+                        Code::H3_ID_ERROR,
+                        format!(
+                            "received a GoAway({}) greater than the former one ({})",
+                            id, prev_id
+                        ),
+                    ));
+                }
+            }
+            *recv_closing = Some(id.into());
+            Ok(())
```
Ruben2424, line 471: Thanks for this awesome PR.

I just got one question, do we need a additional `self.shared.write("connection goaway write").closing = true;` here?

FlorianUekermann: No we don't. This is done on frame receipt on connection level in poll_control (same file L378). That prevented duplication before I merged the client and server side code in the generic `process_goaway`. Maybe I should move it to the place you suggested to keep those things closer together like it is done on shutdown. Let me know what you think. I'm slightly in favor of moving it.

h3/src/client.rs
```diff
@@ -420,22 +429,18 @@ where
...
-                            self.inner.shared.write("poll_close error").error = e.clone().into();
-                            return Poll::Ready(Err(e));
+                            self.inner.shared.write("poll_close error").error = Some(e.clone());
+                            e
                         }
+                    };
+                    if connection_error.is_closed() {
+                        return Poll::Ready(Ok(()));
                     }
+                    return Poll::Ready(Err(connection_error));
```
FlorianUekermann, line 443: I think: This fixes a preexisting bug, which wasn't discovered due to the lack of test coverage of client shutdown (cases where the server closes the connection without error). Basically, if the server closes, the first time the resulting error on the client is seen, the old code does the match on the previous error in `self.inner.shared.error` (which is `None`), thus skipping the arm that leads to `Ok(())`.

The only way I could make this make sense to me was that the intention of this piece of code is to match on the first error, but got the order of doing things confused.
So I did the equivalent to `let connection_error = self.inner.shared.write("poll_close").error.get_or_insert(e)` but without taking the write lock if not necessary.

Did I misunderstand the original intent here?

h3/src/proto/frame.rs
```diff
@@ -30,19 +32,20 @@ impl fmt::Display for FrameError {
...
 pub enum Frame<B> {
     Data(B),
     Headers(Bytes),
-    CancelPush(StreamId),
+    CancelPush(PushId),
     Settings(Settings),
     PushPromise(PushPromise),
-    Goaway(StreamId),
-    MaxPushId(StreamId),
+    Goaway(VarInt),
```
FlorianUekermann, line 47: Just a note: Initially I tried to go for `Frame::Goaway(Goaway)`, where `Goaway` is an enum with `Server` and `Client` variants, as you suggested initially. From a type design perspective I still think that is a little better. But that introduces a dependency on the information whether we are parsing a server or client frame in the frame parser, which in turn infects a few functions and types further up the call-chain, which didn't feel worth it.

h3/src/connection.rs
Ruben2424: Oh I missed that. 
But I am also thinking that it could be slightly better to move it.

h3/src/proto/frame.rs
Ruben2424: Thats ok for me.

h3/src/client.rs
Ruben2424: I do not know what the original intention was. But the change makes sense to me.

seanmonstar: > the lack of test coverage of client shutdown

Thanks for helping to add some, more tests is love! <3

h3/src/connection.rs
FlorianUekermann: done

Issue #144, opened 2022-11-08
Upgrade `quinn` to 0.9.0 for `h3-quinn`


inflation:
With the new release of `quinn`, the APIs are more unified towards a future-based interface. #55 would be a non-issue since the new release has an async `write` function directly.

Note that all futures returned by `quinn` are requiring a lifetime now, storing them may not be an option.

eagr:
Mind if I take this?

inflation:
@eagr Go ahead

seanmonstar:
I recently came to this realization of how we can wrap `async fn`s in poll functions. The benefit of them being poll functions in `h3` is that we can "wait" for many different events at the same time. `async fn`s tend to take a "lock" on the object, especially if it is `&mut self`. But since quinn's types can be cloned, it's possible to do an `async move` dance.

An example is [tokio_util::sync::PollSender](https://docs.rs/tokio-util/latest/src/tokio_util/sync/mpsc.rs.html).

In short, you have an `Option` field holding the quinn struct, and a `tokio_util::syn::ReusableBoxFuture` holding the future. The future moves the struct in, and back out:

```rust
let rx = self.rx.take().unwrap();
self.recv_stream_fut.set(async move {
    let stream = rx.recv_stream().await;
    (rx, stream)
});
```

Once you've polled that future, and gotten the quinn struct and the stream back, you put the struct back into the option, and tada. you have a poll fn.

The h3-quinn crate could have this pattern, in order to implement the `poll_*` functions on the trait. `h3` would not need to know anything about it.

Ralith:
Passing ownership into an aysnc block like that is the approach we recommend, yeah; e.g. the libp2p folks have had some success with it. Other Quinn users have used e.g. [`futures::stream::unfold`](https://docs.rs/futures/0.3.27/futures/stream/fn.unfold.html) to construct `'static` `futures::Stream`s, which saves repeatedly moving the struct in/out.

PR #182, opened 2023-03-28
Quinn 0.9 - quick and dirty experiment with `unfold`

inflation:
As @Ralith said in #144, I hacked a solution using `futures::stream::unfold`. It's not optimized and maybe broken in more ways than expected. But I did get through tests locally. Just put it here for discussion and try my luck on CI.


seanmonstar:
This looks pretty clean, actually! What's the new MSRV required by Quinn?

inflation:
> What's the new MSRV required by Quinn?

1.59, as CI reported.

djc:
@seanmonstar what's your MSRV target/policy?

Issue #177, opened 2023-03-17
Clarify & simplify quic traits


FlorianUekermann:
### Big picture problem

I believe there are a few issues with the `h3::quic::Connection` trait, which lead to expected behavior being under-defined and some hidden bugs in h3, which don't surface in tests due to h3-quinn implementation decisions. The topic easily gets out of hand, because the space of possible interpretations of "correct", as well as the solution space is large, so I'll break it up a bit and leave some stuff out.

#### Example
Possibly a bug, subject to interpretation:
https://github.com/hyperium/h3/blob/da29aea305d61146664189346b3718458cb9f4d6/h3/src/connection.rs#L244-L257
The `Poll::Ready(None)` arm implies that a connection closure is not expected when this is called. Yet, this snippet is actually almost certainly where connection close is observed for the first time. But h3-quinn treats any close as an error, even non-error ones. Therefore the arm that is selected is actually the `Poll::Ready(Err(_))` inside the `?` in L245, which is why no tests have caught this.

If you use other quic implementations, which return `None` on non-error closes, graceful shutdown tests start failing. I think this is a tip-of-the-iceberg situation, because I observe other shutdown related oddities, depending on similar details, which I don't want to get into here for brevity.
Even if you do know what h3 actually expects, I believe there are race conditions, for example: Observing control stream reset is never expected , but implied by a quic connection close.

### Underlying problem
These methods are the meat of the quic traits.
https://github.com/hyperium/h3/blob/da29aea305d61146664189346b3718458cb9f4d6/h3/src/quic.rs#L45-L71
Especially taking into account the comments explicitly stating that `None` is a way to indicate a close. There is a good amount of ambiguity in where connection closes should be communicated first and how. From a naive reading, even a first error surfacing via a RecvStream method seems reasonable, growing the list of affected methods even more.
With the current implementation, there doesn't seem to be a clear distinction between non-error closes and other terminations in the quic traits, which should be handled differently.
I think the example demonstrates that these issues not only make life hard for implementers of the traits, but also for contributors to h3.

### Small detour to increase the solution space
Neither Quic nor Http3 provide a way to accept/reject opening streams/requests by the remote. In Http3 the stream can be closed and stopped immediately with an appropriate error code to indicate rejection after the fact, but opening a stream/request is a unilateral decision by the remote.
[...]

Issue #101, opened 2022-06-24
Integration tests failed on macOS


inflation:
While working on #98, I noticed that two of the integration tests failed on macOS, namely `connect` and `accept_request_end_on_client_close` from `h3-tests/connect.rs`. They both failed on error code 256 (0x100), which means connection closed without a reason in the HTTP/3 RFC. 

Further investigation showed that they passed without a problem in a docker container. Moreover, `quinn`'s tests passed without an error, so it's maybe this crate's doing. Other than that, sometimes they passed when re-run the tests, seems like an I/O or threading issue to me.

Backtrace:
```
---- connect stdout ----
thread 'connect' panicked at 'called `Result::unwrap()` on an `Err` value: h3::Error { kind: ConnectionError(ApplicationClosed(ApplicationClose { error_code: 256, reason: b"" })), code: : Some(256) }', tests/h3-tests/tests/connection.rs:36:53
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1785:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1078:23
   4: connection::connect::{{closure}}::{{closure}}
             at ./tests/connection.rs:36:17
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/future/mod.rs:91:19
   6: <tokio::future::maybe_done::MaybeDone<Fut> as core::future::future::Future>::poll
             at /Users/inflation/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.19.2/src/future/maybe_done.rs:68:48
   7: connection::connect::{{closure}}::{{closure}}
             at /Users/inflation/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.19.2/src/macros/join.rs:116:24
   8: <tokio::future::poll_fn::PollFn<F> as core::future::future::Future>::poll
[...]

PR #183, opened 2023-03-29
Webtransport

darioalessandro:
Hi friends:

@ten3roberts and I are taking on adding this feature, we are not sure if we will keep it here or move it to a separate crate, but this works!!

I already added webtransport support to web apps written in rust via wasm-bindgen: https://github.com/rustwasm/wasm-bindgen/pull/3344

Reddit demanded it: https://www.reddit.com/r/rust/comments/11rfes0/comment/jc8kyax/?utm_source=share&utm_medium=web2x&context=3

**Work left** 
- [x] Move webtransport into its own crate
- [x]  Server initiated Datagrams
- [x]  client initiated Datagrams
- [x]  Better api for incoming bi streams
- [ ]  Session termination
- [x]  Server initiated uni streams
- [x]  Server initiated bi streams
- [ ] Request pass through (possibly as part of incoming bi streams)
- [ ] Rebase onto latest changes


ten3roberts:
Hi again.

We were just in a discussion about forking and implementing webtransport per #71 here over at https://github.com/AmbientRun

What are your plans moving forwards and for the API?

I would like to help with this.

Dropping the webtransport protocol draft for good measure

https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/

Furthermore, there is an implementation in python and go already, so may serve as inspiration.

python sample: https://github.com/GoogleChrome/samples/blob/gh-pages/webtransport/webtransport_server.py
go: https://github.com/adriancable/webtransport-go-example

darioalessandro:
Hey @ten3roberts !!! 

Let's do this! 

I am familiar with aioquic, I came from that lib but had to abandon it because I encountered many leaks, but you are right, as of now it is kind of the "reference" implementation for server side webtransport.

So how about this:

Implement the bare minimal changes in this MR so that we end up with a webtransport server that works as an echo server and we can call from https://security-union.github.io/yew-webtransport/ Yew WebTransport client.

Right now I am looking at the new frame types that we have to add, tonight (Eastern time) I will break this down into tasks, we can tag team this project if you are interested 😄

seanmonstar:
Thanks for starting this effort! I haven't kept track of the spec, so I wanted to ask: how much if it is a protocol *on top of* http3? Is it like websockets? Where it just uses some generic mechanism and otherwise is implemented on top (and this could be a separate library)? Maybe just a new setting, like we did with "extended connect" in h2? Or does it need a lot of integration?

darioalessandro:
Hey @seanmonstar it is extremely similar to websockets.

regarding settings, yes!! there are a couple new settings that the standard introduces

ten3roberts:
My idea is to have a `WebTransportSession` type which is constructed using an `h3::Connection` and established the wt negotiation, similar to the go example listed earlier.

After awaiting the `WebTransportSession` it is used to await incoming streams, datagrams, and open new streams, very much like a `quinn::Connection`

Thoughts?

@darioalessandro I could work on the uni and bi streams of the protocol. Are you going to add me as a contributor to your fork or what is the best way to go about it?

ten3roberts:
On a further note: Should we expose a `Config` struct similar to what quiche does https://docs.rs/quiche/latest/quiche/h3/struct.Config.html to make the webtransport support optional, as all servers using h3 don't want to automatically broadcast and respond that yes, webtransport is indeed supported by the server.

This goes on with my previous comment about the session struct, rather than baking *WebTransport` events into the other existing http/3 events such as requests.

examples/Cargo.toml
```diff
@@ -19,7 +19,7 @@ rustls-native-certs = "0.6"
 structopt = "0.3"
 tokio = { version = "1.6", features = ["full"] }
 tracing = "0.1.10"
-tracing-subscriber = { version = "0.3", default-features = false, features = ["fmt", "ansi", "env-filter", "time", "tracing-log"] }
+tracing-subscriber = { version = "0.3", default-features = true, features = ["fmt", "ansi", "env-filter", "time", "tracing-log"] }
```
darioalessandro, line 22: useless change

examples/generate_certs.sh
```diff
@@ -0,0 +1,7 @@
+
+# use mkcert to generate certs for localhost
+mkcert localhost
```
darioalessandro, line 3: This is unrelated to our work, just makes it easier for people to get certs that are trusted by chrome

examples/server.rs
```diff
@@ -54,15 +54,13 @@ pub struct Certs {
...
-static ALPN: &[u8] = b"h3";
-
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     tracing_subscriber::fmt()
         .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
         .with_span_events(tracing_subscriber::fmt::format::FmtSpan::FULL)
         .with_writer(std::io::stderr)
-        .with_max_level(tracing::Level::INFO)
+        .with_max_level(tracing::Level::TRACE)
```
darioalessandro, line 63: Increased logging level so that we can see the quic frames

h3/src/connection.rs
```diff
@@ -109,6 +109,12 @@ where
         settings
             .insert(SettingId::MAX_HEADER_LIST_SIZE, max_field_section_size)
             .map_err(|e| Code::H3_INTERNAL_ERROR.with_cause(e))?;
+        settings.insert(SettingId::ENABLE_CONNECT_PROTOCOL, 1)
```
darioalessandro, line 112: attempt to send the webtransport headers on the server side

h3/src/server.rs
```diff
@@ -400,7 +403,7 @@ where
     fn poll_control(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Error>> {
         while let Poll::Ready(frame) = self.inner.poll_control(cx)? {
             match frame {
-                Frame::Settings(_) => trace!("Got settings"),
+                Frame::Settings(w) => trace!("Got settings > {:?}", w),
```
darioalessandro, line 406: here I want to intercept when the browser sent the enable_webtransport setting

darioalessandro:
@ten3roberts I added you as maintainer. 

1. Regarding config: how about adding a second parameter to h3::server::Connection::new with the config struct (to your point, similar to quiche), right now, it just takes que quinn connection: https://github.com/hyperium/h3/blob/3bb88ab369df5cd1fe0c2b498f772bc556d47ccb/h3-quinn/src/lib.rs#L40
2. Do you want to get started with this? how should we divy up the work?

ten3roberts:
Thanks, you too.

1. That would be a very good idea, since the builder way to create a connection is less obvious/hidden, and being clear about what exists is usually good.

2. I could work with anything really. We will need support for both the server and client part of the things. What do you think about my `WebTransportSession` type proposal, making a webtransport opt in? First order of business would be establishing a session on either side.

PS: My timezone is Z+0200 and will be working days between roughly 9am-5am
I'll fix the merge conflicts on my PR and then you can merge the PR and we can work on the same branch

darioalessandro:
@ten3roberts 
1. move forward with the `WebTransportSession` approach sounds good to me. 
2. With regards to the client, you can test with my ui https://security-union.github.io/yew-webtransport/ it is not perfect so do keep the console open (sometimes it does not show all errors)

h3/src/server.rs
ten3roberts: I am working on that now

Ruben2424:
Hi,

I don't want to mix everything up here, but I would like to share my thoughts.

I think it would be best to only implement http/3 in the h3 crate itself without extensions and only implement necessary parts to support extensions implemented in a separate crate (maybe `extended connect` I think @eagr started working on this in #159 but I don't know what the status is).
And to provide a API in order to be able to implement extensions to the protocol like WebTransport ontop. 
Maybe some functionality to send Settings, Streams, etc. that the h3 don’t know and a way to receive these unknown types. 

For WebTransport there are 2 Protocols needed. Http-Datagrams and WebTransport itself. I think both could be in a separate crate each. 

I didn’t thought about this too much, so I maybe didn’t see some obvious problems with this proposal or there might be a better solution.

examples/generate_certs.sh
darioalessandro: @ten3roberts this script will help to generate certs that Google Chrome trusts, https://github.com/FiloSottile/mkcert If you are on MacOS you can `brew install mkcert`

inflation:
Well, I think if there's another protocol upcoming, then that's the time we'd separate things and make a general extension mechanism. For now it's ok to include WebTransport but maybe behind a feature flag.

darioalessandro:
Hey @Ruben2424 @inflation !! 

Thank you so much for your feedback. 

I discussed with @ten3roberts and we are happy to move this to a different crate after we get it to work and just including the bare minimal in this repo.

also, thank you for mentioning #159 !

darioalessandro:
@ten3roberts got the connect to work by adding the appropriate headers to the response, now we need to listen to inbound datagrams, bidirectional and unidirectional streams. 

I had to switch to the sample webtransport client because the error handling is better than in my yew-webtransport and I can actually see the errors in the browser as opposed to weird wasm traces:

https://webtransport.day/   
Screen Shot 2023-03-31 at 11 29 15 PM

darioalessandro:
@ten3roberts I had to hardcode a few things and break encapsulation to get the thing to work, but it is done. 

Now we can polish it with a nice architecture!! 

https://user-images.githubusercontent.com/1176339/229327195-d6a7f8b8-1e6a-4bf4-a375-c98959f0cd72.mov

ten3roberts:
Attempting to cut through some of the noise in this thread caused by my comments.

@seanmonstar is it ok if I add the [pin-project](https://crates.io/crates/pin-project) crate as a dependency.

I need to create some stateful futures which hold their own internal state, rather than mutating state inside the "parent" like connection. This makes `poll_` style functions harder to write as they do not retain intermediate state and are thus lossy for some use cases.

I also plan on reverting the `pub(crate)` I added to the `Connection` by instead adding the appropriate methods and abstractions rather than reaching into the internals to grab what we need, which is a dirty trick.

seanmonstar:
> is it ok if I add the pin-project crate as a dependency?

Is it possible to use pin-project-lite? It works for most of the cases, and compiles much faster.

ten3roberts:
Should the `webtransport::RecvStream` and coming `webtransport::SendStream` implement `futures_io::AsyncRead`/`futures_util::AsyncWrite`, or should only `quic::RecvStream` be implemented, and reading/integration with buffered readers be left as an exercise to the user using `RecvStream::poll_data`?

darioalessandro:
@ten3roberts what's left? 

Can you please create a list? I would like to help too :)

ten3roberts:
It course

- [ ] Better api for incoming bi streams
- [ ] Session termination
- [ ] Server initiated uni streams
- [ ] Server initiated bi streams
- [ ] Request pass through (possibly as part of incoming bi streams)
- [ ] Rebase onto latest changes

For maintainers

I implemented the proposed API change in #78 which allows sending raw bytes rather than just encoded frames which was required for streaming data in the streams, rather than sending complete frames. Frame encoding is still supported as `WriteBuf` implements `Buf`, so the same functionally but more composable.

I also lowered the trait generic of the stream into the `poll_send` method which made the whole api a lot easier to use and less cluttery with bounds and PhantomData. This was already half way done as the previous send method was *also* generic over the `Into<WriteBuf>`, but as said, is now more flexible in terms of *what* you can use to send your data.

See the commit message of https://github.com/hyperium/h3/commit/16eead4aa95551087cd60ac1da1e1f00fc2c99b4 for more details.

Please do let me know what you think of that decision, of it there is any other way to go around it

darioalessandro:
@ten3roberts receiving unidirectional streams works great. 

I think there's an issue with bi-directional streams, seems like the server reads only the first chunk.

ten3roberts:
Alright will look into it.

We also may need to add a webtransport `Client`, though I am not sure if it is in the scope for this PR, most of the groundwork is there as it shares a lot with the server

ten3roberts:
The test `tests::connection::control_close_send_error` fails on master and persists to this day on d7a237f539b4c153318004283d776ff6724db830.

Is this something you are aware of?

darioalessandro:
> Alright will look into it.
> 
> We also may need to add a webtransport `Client`, though I am not sure if it is in the scope for this PR, most of the groundwork is there as it shares a lot with the server

I think that by adding a test webtransport client we could add some sweet integration tests!!!

darioalessandro:
> The test `tests::connection::control_close_send_error` fails on master and persists to this day on [d7a237f](https://github.com/hyperium/h3/commit/d7a237f539b4c153318004283d776ff6724db830).
> 
> Is this something you are aware of?

@Ruben2424 do you know the answer to this?

inflation:
> The test `tests::connection::control_close_send_error` fails on master and persists to this day on [d7a237f](https://github.com/hyperium/h3/commit/d7a237f539b4c153318004283d776ff6724db830).
> 
> Is this something you are aware of?

It's a known error that only happens on macOS #101. I don't have a clear answer about why it happens.

darioalessandro:
@ten3roberts I'll take on:

 Server initiated uni streams
 Server initiated bi streams
 
today

ten3roberts:
Fantastic.

I've already started a bit on the bi streams. The uni streams should be almost identical except the type.

Opening is a lot easier than incoming since you don't need to do partial framed decoding.

darioalessandro:
Hey @ten3roberts it's been a pretty frustrating session for me. 

Here's the progress:

1. I stabilized the webtransport_server.rs script, I commented server initiated uni streams for now because somehow that cause the connection to close. 

2. I added receiving server initiated uni and bidi streams on the web client: https://github.com/security-union/yew-webtransport/pull/6

If you want to use yew-webtransport make sure to checkout the readme. 

3. I seem to be able to send the server initiated uni stream to the client here:
https://github.com/hyperium/h3/blob/1d8339df9b9abf029b9cfa21a34865f6ab4c54b5/examples/webtransport_server.rs#L273-L276

But the client never receives the uni_stream here:
https://github.com/security-union/yew-webtransport/blob/2b2182d2341cf152b34401a0a23881f3c06448c4/src/webtransport.rs#L163

I use this same pattern for datagrams, so I have no idea why this is not working for unistreams: 
https://github.com/security-union/yew-webtransport/blob/2b2182d2341cf152b34401a0a23881f3c06448c4/src/webtransport.rs#L203 

Let me know what you think.

darioalessandro:
I am confident that I am missing the stream type and session id:

image

darioalessandro:
Stealing from quiche: https://github.com/security-union/webtransport-quiche/blob/05d7d3a2b5ea8d512f00ea719259353c84bbeb4f/src/lib.rs#L249-L271

examples/webtransport_server.rs
```diff
@@ -0,0 +1,328 @@
...
+
+                // Send session id
+                {
+                    let mut data = [0u8; 8];
+                    let mut buf = octets::OctetsMut::with_slice(&mut data);
+                    buf.put_varint(session_id.0).unwrap();
+                    send.write(&mut buf.buf()).await.context("Failed to respond")?;
+                }
+                // Send actual data.
+                send.write(&mut message).await.context("Failed to respond")?;
```
darioalessandro, line 298: @ten3roberts can you spot what I am doing wrong here? 

I am trying to send the stream type followed by the session id, then the data per the spec.

This is the quiche equivalent for sending the stream type: https://github.com/cloudflare/quiche/blob/330b24e2f01ac4e16313154f753a9228cd9e1061/quiche/src/h3/mod.rs#L2082-L2120

examples/launch_chrome.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
darioalessandro, line 1: This can bed it's own MR if we want to reduce line count, it is really useful so imo should go in either way.

h3/src/connection.rs
```diff
@@ -156,9 +218,11 @@ where
         //# Endpoints MUST NOT require any data to be received from
         //# the peer prior to sending the SETTINGS frame; settings MUST be sent
         //# as soon as the transport is ready to send data.
-        stream::write(
+        trace!("Sending Settings frame: {settings:#x?}");
+        stream::write::<_, _, Bytes>(
             &mut control_send,
-            (StreamType::CONTROL, Frame::Settings(settings)),
+            // (StreamType::CONTROL, Frame::Settings(settings)),
```
darioalessandro, line 224: remove commented code

h3/src/connection.rs
```diff
@@ -311,28 +395,27 @@ where
...
         }
 
-        loop {
-            match self.poll_accept_recv(cx) {
-                Poll::Ready(Ok(_)) => continue,
-                Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
-                Poll::Pending if self.control_recv.is_none() => return Poll::Pending,
-                _ => break,
+        let recv = {
+            // TODO
```
darioalessandro, line 408: what's left here @ten3roberts ?

h3/src/webtransport/server.rs
```diff
@@ -0,0 +1,427 @@
...
+{
+    /// Accepts a *CONNECT* request for establishing a WebTransport session.
+    ///
+    /// TODO: is the API or the user responsible for validating the CONNECT request?
+    pub async fn accept(
+        request: Request<()>,
+        mut stream: RequestStream<C::BidiStream>,
+        mut conn: Connection<C>,
+    ) -> Result<Self, Error> {
+        // future::poll_fn(|cx| conn.poll_control(cx)).await?;
```
darioalessandro, line 56: TODO: remove commented code.

h3/src/webtransport/server.rs
```diff
@@ -0,0 +1,427 @@
...
+    C: quic::Connection,
+{
+    type Output = Result<Option<(SessionId, Bytes)>, Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        tracing::trace!("poll: read_datagram");
+
+        let mut conn = self.conn.lock().unwrap();
+        match ready!(conn.poll_accept_datagram(cx))? {
+            // Some(v) => Poll::Ready(Ok(Some(Datagram::decode(v)?.payload))),
```
darioalessandro, line 381: TODO: remove this

darioalessandro:
@ten3roberts you can scratch server initiated bi streams!!! those work now!

h3/src/proto/stream.rs
```diff
@@ -26,6 +26,7 @@ stream_types! {
     PUSH = 0x01,
     ENCODER = 0x02,
     DECODER = 0x03,
+    WEBTRANSPORT_BIDI = 0x41,
```
ten3roberts, line 29: I do not think that bi streams types should be here, as (from my understanding) this struct if for unidirectional stream types as defined in https://www.rfc-editor.org/rfc/rfc9114#section-11.2.4, whereas the `WEBTRANSPORT_BIDI` is a frame type for the HTTP/3 framing layer

darioalessandro: Ok, happy no to move it, where would you put it?

seanmonstar:
Feedback welcome in this discussion about where the code should live: https://github.com/hyperium/h3/discussions/189

h3-webtransport/src/server.rs
```diff
@@ -0,0 +1,422 @@
...
+                // We always send a last GoAway frame to the client, so it knows which was the last
+                // non-rejected request.
+                // self.shutdown(0).await?;
+                todo!("shutdown");
+                // return Ok(None);
+            }
+            Err(err) => {
+                match err.inner.kind {
+                    h3::error::Kind::Closed => return Ok(None),
+                    // h3::error::Kind::Application {
```
darioalessandro, line 177: @ten3roberts had to comment this out because I am getting: 

```
error[E0638]: `..` required with variant marked as non-exhaustive
   --> h3-webtransport/src/server.rs:177:21
    |
177 | /                     h3::error::Kind::Application {
178 | |                         code,
179 | |                         reason,
180 | |                         level: ErrorLevel::ConnectionError,
181 | |                     } => {
    | |_____________________^
```

h3-webtransport/src/server.rs
```diff
@@ -0,0 +1,422 @@
...
+        ReadDatagram {
+            conn: self.conn.lock().unwrap().inner.conn.clone(),
+        }
+    }
+
+    /// Sends a datagram
+    ///
+    /// TODO: maybe make async. `quinn` does not require an async send
+    pub fn send_datagram(&self, data: impl Buf) -> Result<(), Error> {
+        self.conn
```
darioalessandro, line 140: @seanmonstar we had to expose the h3 connection via Arc<Mutex> what do you think about this?

darioalessandro:
@ten3roberts I reverted multi session for now, and moved web transport to its own crate, please feel free to continue to move things out of the h3 crate and into webtransport, right now it seems to me that h3 and webtransport are tightly coupled.

h3/src/lib.rs
```diff
@@ -12,21 +12,37 @@ pub use error::Error;
 pub use proto::headers::Protocol;
 
 mod buf;
+
+#[cfg(feature="allow_access_to_core")]
```
darioalessandro, line 16: This is my attempt to make it hard to expose the core modules that webtransport need, let me know what you think

ten3roberts: Perhaps the conventional export, and then a separate module, called `internals` which is conditonal and re-exports all private imports

.github/workflows/CI.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         with:
           command: clippy
 
-  msrv:
+  gsrv:
```
seanmonstar: Why the change from m to g?

h3/src/error.rs
```diff
@@ -12,7 +12,8 @@ pub(crate) type TransportError = Box<dyn quic::Error>;
 /// A general error that can occur when handling the HTTP/3 protocol.
 #[derive(Clone)]
 pub struct Error {
-    pub(crate) inner: Box<ErrorImpl>,
+    /// The error kind.
+    pub inner: Box<ErrorImpl>,
```
seanmonstar, line 16: Could this be kept private? And the struct below?

h3/src/lib.rs
```diff
@@ -5,17 +5,25 @@
...
 pub mod server;
 
 pub use error::Error;
+pub use proto::headers::Protocol;
 
 mod buf;
-mod connection;
-mod frame;
-mod proto;
+#[allow(missing_docs)]
```
seanmonstar, line 17: How much needs to be made public outside the crate? This makes a lot of the library internals public, which means refactors would be "breaking changes".

h3/src/proto/frame.rs
```diff
@@ -62,11 +75,21 @@ impl From<usize> for PayloadLen {
 }
 
 impl Frame<PayloadLen> {
-    pub const MAX_ENCODED_SIZE: usize = VarInt::MAX_SIZE * 3;
+    pub const MAX_ENCODED_SIZE: usize = VarInt::MAX_SIZE * 7;
```
seanmonstar, line 78: This change might be obvious, sorry to ask, but this seems like a big jump. Is it spec related? (I actually have completely forgotten what this associated constant is really for...)

h3/src/proto/frame.rs
```diff
@@ -389,9 +431,22 @@ setting_identifiers! {
...
+    H3_DATAGRAM = 0xFFD277,
+    // https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-8.2
+    ENABLE_WEBTRANSPORT = 0x2B603742,
+    // https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3/#section-8.2
+    H3_SETTING_ENABLE_DATAGRAM_CHROME_SPECIFIC= 0xFFD277,
+
+    WEBTRANSPORT_MAX_SESSIONS = 0x2b603743,
+    // Dummy setting to check it is correctly ignored by the peer.
+    // https://datatracker.ietf.org/doc/html/rfc9114#section-7.2.4.1
+    DUMMY = 0x21,
```
seanmonstar, line 446: This is like the grease frames we send and receive, right? #92

h3/src/proto/headers.rs
```diff
@@ -77,6 +86,7 @@ impl Header {
         //# If the scheme does not have a mandatory authority component and none
         //# is provided in the request target, the request MUST NOT contain the
         //# :authority pseudo-header or Host header fields.
+        trace!("got headers {:#?}", self.fields);
```
seanmonstar, line 89: I would recommend against logging the fields, this has caused problems in hyper (and plenty of internal privacy reviews of any language/framework), it's too easy to accidentally log sensitive information.

h3/src/quic.rs
```diff
@@ -29,16 +30,59 @@ impl<'a, E: Error + 'a> From<E> for Box<dyn Error + 'a> {
...
+        match self {
+            Self::ConnectionLost(err) => err.err_code(),
+            _ => None,
+        }
+    }
+}
+
 /// Trait representing a QUIC connection.
-pub trait Connection<B: Buf> {
+pub trait Connection {
```
seanmonstar, line 73: So far, we've allowed implementations/users to specify a different `Buf` type than `Bytes`, in case they want to make better use of vectored writes, or just less copies in general. Seems like a good property to keep.

h3/src/quic.rs
```diff
@@ -29,16 +30,59 @@ impl<'a, E: Error + 'a> From<E> for Box<dyn Error + 'a> {
...
     /// The type produced by `poll_accept_bidi()`
-    type BidiStream: SendStream<B> + RecvStream;
+    type BidiStream: BidiStream<RecvStream = Self::RecvStream, SendStream = Self::SendStream>;
```
seanmonstar, line 75: By originally keeping it only bound to `SendStream + RecvStream`, and the `trait BidiStream` being optional, this made it so implementations could opt-in to being splittable.

h3/src/lib.rs
darioalessandro: I tried to only make public the bare minimal. 

users would have to go out of their way to enable the feature "allow_access_to_core" to enable access to all the private crates. 

it reminds me of rustls `dangerous_configuration` I can actually just call it dangerous config. 

https://github.com/rustls/rustls/blob/3d8b801221a6b4adfdca04934ecdda497423a0d8/rustls/Cargo.toml#L27

h3/src/lib.rs
darioalessandro: I like this!!! what do you think @seanmonstar ?

h3/src/proto/headers.rs
darioalessandro: makes sense

h3/src/proto/frame.rs
ten3roberts: It dictates the size of the WriteBuf which is used to encode the settings frame.

The previous limit was too small and panicked.

We need to send: `H3_DATAGRAM`, `H3_ENABLE_WEBTRANSPORT`, `ENABLE_CONNECT_PROTOCOL`, and `WEBTRANSPORT_SESSIONS` ontop of what is already sent in the settings frame

h3/src/quic.rs
ten3roberts: Alright, seems like a good idea.

Will add combined webtransport bidi streams

ten3roberts: Done

ten3roberts: Done

h3/src/connection.rs
redoriental: I would like to know the release time of the webtransport server. I am really looking forward to your project

darioalessandro: Hey @redoriental I encourage you to try out the h3-webtransport crate by installing it from this branch using:

h3-webtransport = { git= "https://github.com/security-union/h3.git", rev = "f4e83fb9be28ef3ea99e5feed725550f0169e5c2" } that way you wont have to wait for this PR to be merged. 

We are currently working on separating h3 and h3-webtransport, and based on the ongoing discussions, it appears that this process may take some time.

The nature of open-source development can result in a slower pace, as contributors often have limited time to dedicate, and individuals from diverse backgrounds and cultures must come to a consensus on various objectives. However, rest assured that we will ultimately deliver a well-organized and efficient webtransport server!

h3/src/proto/frame.rs
darioalessandro: Looking now

darioalessandro: Dead code, probably I took it from aioquic but did not end up using it, I'm going to delete this.

h3/src/quic.rs
ten3roberts: If I've understood it correctly:

You want to allow

```rust
impl Connection<Vec<u8>> for MyConn { ... }
impl Connection<Bytes> for MyConn { ... }
impl Connection<Vec<Bytes>> for MyConn { ... }
```

Rather than a blanketed impl for *all* Buf types like quinn and s2n does?

In that case, perhaps there is a better way to allow this which doesn't constrain the user in what data they can send, or doesn't leak the underlying buffering type all the way up the the user API.

seanmonstar: Sort of, yea. More like in hyper and h2, they're just `Connection<Io, BufType>`, and whatever kind of the buf the user has, we can work with it.

seanmonstar:
Zooming out a little, I'd like to suggest some targets for what `h3` would expose:

- As little change as possible to `h3::quic` traits. If the existing traits currently _prevent_ something webtransport needs, we can discuss.
- Expose as little of the `h3` internals as possible. Ideally, the only public modules would be `client`, `server`, and `quic`. Maybe another module is needed for extensions?
- The exported type from `h3` that is basically a webtransport could potentially just have inherent methods, and then `h3-webtransport` uses them directly. Then `h3` doesn't have to pick which traits to implement for its exported types, and `h3-webtransport` doesn't care either.

Of course, those are generalities. We can be flexible if there's need to be.

Does that help get across what I was commenting on in my previous review?

h3/src/quic.rs
```diff
@@ -75,14 +119,23 @@ pub trait Connection<B: Buf> {
...
+    /// Send a datagram
+    fn send_datagram(&mut self, data: Bytes) -> Result<(), SendDatagramError>;
 }
 
 /// Trait for opening outgoing streams
-pub trait OpenStreams<B: Buf> {
+pub trait OpenStreams {
     /// The type produced by `poll_open_bidi()`
-    type BidiStream: SendStream<B> + RecvStream;
+    type BidiStream: BidiStream;
```
ten3roberts, line 136: Realized this should be SendStream + RecvStream again

h3/src/quic.rs
```diff
@@ -122,15 +182,15 @@ pub trait SendStream<B: Buf> {
     fn reset(&mut self, reset_code: u64);
 
     /// Get QUIC send stream id
-    fn id(&self) -> StreamId;
+    fn send_id(&self) -> StreamId;
```
ten3roberts, line 185: Avoids disambiguity with `id` on `RecvStream` as a send stream id if required for the webtransport control stream to identify the session

darioalessandro:
@seanmonstar @Ruben2424 i think that @ten3roberts adjusted the code a bit.

Can we have another PR review? 

Please feel free to do some of the code changes so that this PR better fits your crate.

I am worried that this is getting stale and I would really like to see it in prod. 

🙌🙌👍🤗🤗

ten3roberts:
Yes, I did change the code according to the comments, most notably restore the quic trait and break out the unframed sending.

As Dario said: feel free to change or suggest things for the code to better align with h3's vision.

Ruben2424:
Hi @darioalessandro ,

I hope you're doing well. I want to thank you and @ten3roberts for the effort you've put into this.

I want to assure you that I haven't forgotten about the work. Unfortunately, I didn't have time to review it in the past few weeks. However, I'm very excited about this feature, and I hope to find some time soon to take a closer look.

daxpedda:
Firefox decided to ship WebTransport in v114 and enabled it on nightly by default: https://bugzilla.mozilla.org/show_bug.cgi?id=1818754.

Would be nice if this is also tested against Firefox, just to get more coverage.

h3-quinn/src/lib.rs
```diff
@@ -21,8 +22,12 @@ use futures::{
 pub use quinn::{
     self, crypto::Session, AcceptBi, AcceptUni, Endpoint, OpenBi, OpenUni, VarInt, WriteError,
 };
+use quinn::{ReadDatagram, SendDatagramError};
 
-use h3::quic::{self, Error, StreamId, WriteBuf};
+use h3::{
+    quic::{self, Error, StreamId},
+    stream::WriteBuf,
```
seanmonstar, line 29: Should this export be left in `h3::quic`?

h3/src/lib.rs
seanmonstar: The concern is that people can enable the feature and depend on the internals, and when we change things around in a refactor (such as in this PR), the public stuff changes and "breaks" them.

The "dangerous configuration" in rustls I don't think is meant to access unstable stuff, but rather to access crypto-security things. More like a "woah, are you sure".

Even if used just by `h3-webtransport`, if we make a change in `h3`, anyone depending on `h3-webtransport` who updates their `h3` dependency will see it "break".

h3/src/proto/headers.rs
```diff
@@ -293,6 +311,22 @@ where
     R::from_str(s).map_err(|_| HeaderError::invalid_value(name, value))
 }
 
+#[derive(Copy, PartialEq, Debug, Clone)]
```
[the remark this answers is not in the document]
seanmonstar: We've made a top-level `ext` module in other libraries before, for this kind of thing.

h3/src/quic.rs
```diff
@@ -75,6 +120,15 @@ pub trait Connection<B: Buf> {
 
     /// Close the connection immediately
     fn close(&mut self, code: crate::error::Code, reason: &[u8]);
+
+    /// Poll the connection for incoming datagrams.
+    fn poll_accept_datagram(
```
seanmonstar, line 125: Are these actions "required" for h3 and quic? Or is it an extension, that some implementations might not bother to do? If it's optional, how should we allow people to not implement it?

Maybe as a separate trait, that if also implemented, then people can access the methods on `h3::client` or `h3::server` to deal with datagrams?

h3/src/quic.rs
```diff
@@ -122,15 +176,29 @@ pub trait SendStream<B: Buf> {
...
+    ) -> Poll<Result<usize, Self::Error>>;
 }
 
 /// A trait describing the "receive" actions of a QUIC stream.
 pub trait RecvStream {
     /// The type of `Buf` for data received on this stream.
     type Buf: Buf;
     /// The error type that can occur when receiving data.
-    type Error: Into<Box<dyn Error>>;
+    type Error: 'static + Error;
```
seanmonstar, line 201: Could this bound stay as `Into<Box<...>>`?

h3/src/webtransport/session_id.rs
```diff
@@ -0,0 +1,50 @@
...
+    coding::{Decode, Encode},
+    stream::{InvalidStreamId, StreamId},
+    varint::VarInt,
+};
+
+/// Identifies a WebTransport session
+///
+/// The session id is the same as the stream id of the CONNECT request.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct SessionId(pub u64);
```
seanmonstar, line 13: I recommend that the field should be private.

seanmonstar (changes requested):
Thanks for the continued refinements! I think this is in much better shape, and close to being merged.

Most of my concerns are around keeping the public API as minimal as possible, to allow us to refactor as needed. Comments inline. <3

h3/src/quic.rs
ten3roberts: It is an HTTP/3 extension.

I could quite easily split it into a trait inside of h3/quic, if so desired.

https://www.rfc-editor.org/rfc/rfc9297

h3/src/webtransport/session_id.rs
ten3roberts: Done

h3/src/webtransport/mod.rs
```diff
@@ -0,0 +1,2 @@
+mod session_id;
+pub use session_id::*;
```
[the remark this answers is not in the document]
ten3roberts: Done

h3/src/quic.rs
ten3roberts: I have now split it into a `SendDatagramExt` and `RecvDatagramExt` that use the same "buffer specialization" pattern we've talked about before. Looks a lot cleaner now in my opinion.

h3/src/server.rs
```diff
@@ -467,6 +515,102 @@ where
     }
 }
 
+/// Configures the HTTP/3 connection
+#[derive(Debug, Clone, Copy)]
+pub struct Config {
```
Ruben2424, line 520: At the moment there are 2 ways to create a `Connection`. One to use a this `Config` struct and a second using `Builder` (which uses this struct under the hood). I think there should be only one way to create a new Connection. 

The current `Builder`s have the advantage to be able to support different configurations for client and server if needed.
For the common stuff between client and server is is fine to use this struct. But then is should probably get its own file.

h3/src/connection.rs
```diff
@@ -69,33 +73,76 @@ pub trait ConnectionState {
...
+#[allow(missing_docs)]
 pub struct ConnectionInner<C, B>
 where
     C: quic::Connection<B>,
     B: Buf,
 {
     pub(super) shared: SharedStateRef,
-    conn: C,
+    /// TODO: breaking encapsulation just to see if we can get this to work, will fix before merging
+    pub conn: Arc<Mutex<C>>,
```
Ruben2424, line 106: Maybe we could remove this `Arc<Mutex<C>>`.
Some implementations like quinn allow to cheaply clone the connection using a `Mutex` under the hood. 
So we could just add a bound for `C` and require it to be `Clone`.

If a Connection does not support cheaply cloning it could use a `Arc<Mutex<>>` when implementing the quic connection traits.

h3/src/connection.rs
```diff
@@ -106,11 +153,33 @@ where
...
+            .insert(
+                SettingId::ENABLE_WEBTRANSPORT,
+                config.enable_webtransport as u64,
+            )
             .map_err(|e| Code::H3_INTERNAL_ERROR.with_cause(e))?;
+        settings
+            .insert(SettingId::H3_DATAGRAM, config.enable_datagram as u64)
+            .map_err(|e| Code::H3_INTERNAL_ERROR.with_cause(e))?;
+
+        tracing::debug!("Sending server settings: {settings:#x?}");
```
Ruben2424, line 180: CI says this is breaking MSRV. 
Is this log duplicate with the same in Line 228 of this file?

Ruben2424 (changes requested):
This is Awesome :muscle:, thanks for all the effort.
I left some thoughts inline.

h3/Cargo.toml
```diff
@@ -19,11 +19,15 @@ categories = [
     "web-programming::http-server",
 ]
 
+[features]
+allow_access_to_core = []
```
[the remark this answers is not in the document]
ten3roberts: That's fantastic.

Makes it a lot more clear that this is a no-no feature rather than something that sounds like a good "normal" feature to enable

h3/src/connection.rs
ten3roberts: Oh, didn't see that that change was made by us. Thought it was there from the beginning.

Will see if I can revert the change

h3/src/connection.rs
ten3roberts: It is the usage of inline formatting.

Will fix though running cargo check on my end with 1.57 gives a "failed to select version for `time`"

h3/Cargo.toml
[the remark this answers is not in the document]
darioalessandro: I love this!!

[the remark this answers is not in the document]
ten3roberts: Great. This would still break if we do a breakage in the now exposed backend, if `h3-webtransport` is present in the dependency tree due to cargo feature resolution.

However, `h3-webtransport` *should* usually reside near the top in your local crates and not external crates.io dependencies. So fixing these issues would be far easier.

The feature has now been renamed

h3-webtransport/src/server.rs
```diff
@@ -0,0 +1,421 @@
...
+        })
+        .await;
+
+        let mut stream = match stream {
+            Ok(Some(s)) => FrameStream::new(BufRecvStream::new(s)),
+            Ok(None) => {
+                // We always send a last GoAway frame to the client, so it knows which was the last
+                // non-rejected request.
+                // self.shutdown(0).await?;
+                todo!("shutdown");
```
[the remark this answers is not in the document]
ten3roberts: Good catch.

For the h3/server.s this calls `self.shutdown`.

This is not possible for the webtransport session due to the mutex guarding the server_conn, at shutdown is async.

How do you reckon we should go about this to reject HTTP/3 requests and shutdown the HTTP/3 server?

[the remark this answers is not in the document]
Ruben2424: Hmm. 
Maybe we can just remove the `todo!` and not call shutdown here.
1. I am not sure if this last shutdown is even required by the http3 spec. 
2. And the `None` Option is passed from the quic trait implementations. It is not realy clear for what the `None` stands for (see #177 ) the quinn implementation does not use it at all. 

what do you think @seanmonstar ?

ten3roberts:
@Ruben2424 are there any features or traits, such as `futures::AsyncRead` that should be placed behind feature flags, or is it good as is?

h3/src/server.rs
darioalessandro: I see so the ask here is to continue using the builder pattern as opposed to passing a config object. 

This means that all properties in the config object have to be moved to `Connection`

Ruben2424: Not quite, the `Connection` can still have the `config` object. There was 2 ways to create a connection.
1. Create a `Config` with all necessary properties and construct a `Connection` with this `Config`
2. Use a builder which constructs the `Config` and the `Connection`

@ten3roberts already removed option one in [here](https://github.com/hyperium/h3/pull/183/commits/3cb795546f50961c3bd7030d7015333e4a55525b)
Yet there is only one way to create a `Connection` using the Builder. 
Since the `Client` also uses the `Config` struct, the only thing left is to move `Config` out of the `server.rs` file. IMO it would be best to create a `config.rs` file.

Ruben2424:
> @Ruben2424 are there any features or traits, such as futures::AsyncRead that should be placed behind feature flags, or is it good as is?

I can't thing of a reason to put traits behind a feature flag. But I do not have much experience with those traits so I might missed something. What do you think @seanmonstar

h3/src/server.rs
ten3roberts: If I understand it all corretly you think it is fine to keep the `Config` struct but only use it internally, if that is correct?

I moved the fields inside the builder into a separate `Config` struct, because with all the options for webtransport, connect, datagrams, etc it became cumbersome to pass all those around separately and then have as "inlined" fields in the resulting connection.

ten3roberts:
@Ruben2424 I have now fixed all of your suggestions, moved config to a separate file, and fixed the msrv.

Can we re-run the CI?

Is there anything more or could this be merged? I am going to be available going onwards and will be able to fix and submit PR if any issue arises.

ten3roberts:
Getting an error about section-3.1 not exiting in the duvet report, but the link works.

Not sure if I've missed anything, I tried adding the link to the rfc in the `ci/compliance/extract.sh` but it didn't help

h3/src/connection.rs
ten3roberts: Fixed :)

ten3roberts:
It seems to be as the link is not an RFC editor link. I removed the only offending section link for now.

The CI should now be all green

ten3roberts:
@Ruben2424 or @seanmonstar 

Could we get a final CI and review? Is there anything more to address that I have missed?

Ruben2424 (approved):
Thanks for all the great work!
In my opinion this is ready to merge now. 
I would like to wait with merging until @seanmonstar approved this too.

darioalessandro:
Btw, if you have a specific recommendation to keep imports as private as possible let me know. 

I am also happy to keep webtransport as a module within h3 until we figure out how to split it.

@Ruben2424 @seanmonstar

Ruben2424:
> Btw, if you have a specific recommendation to keep imports as private as possible let me know.

I have some early ideas about creating one or multiple `Extention` traits. Then for Example the h3 `Connection` struct could hold a struct whicht impls the trait. H3 could call methods on this struct on specific events for example when the peer opens a stream. 

Then only the traits and parameter types would need to be public and all of the internal structure could be private. 

But that are just some early ideas from my side that might not work well. 

In my opinion this should be explored after this PR was merged.

darioalessandro:
Sgtm let's go 🙌🙌🙌

kixelated:
I migrated [my project](https://github.com/kixelated/moq-rs) to this PR and it works great. I was using a quiche fork, but things started to get hairy without async support. Thank you for the hard work!

One problem I'm running into is access to the underlying QUIC library. I'd like to use [SendStream.set_priority](https://docs.rs/quinn/latest/quinn/struct.SendStream.html#method.set_priority) and eventually [Connection.congestion_state](https://docs.rs/quinn/latest/quinn/struct.Connection.html#method.congestion_state). These are both implementation-specific, so I don't think adding them to the `h3::quic` traits makes sense. Any suggestions?

Ruben2424:
> One problem I'm running into is access to the underlying QUIC library. I'd like to use [SendStream.set_priority](https://docs.rs/quinn/latest/quinn/struct.SendStream.html#method.set_priority) and eventually [Connection.congestion_state](https://docs.rs/quinn/latest/quinn/struct.Connection.html#method.congestion_state). Any suggestions?

Hi @kixelated , You can try to `clone()` the `quinn::Connection` before constructing the `h3::Connection` and then call  [Connection.congestion_state](https://docs.rs/quinn/latest/quinn/struct.Connection.html#method.congestion_state) on the clone.
I have no idea for the [SendStream.set_priority](https://docs.rs/quinn/latest/quinn/struct.SendStream.html#method.set_priority) right now. But fell free to open a issue to further discuss this.

kixelated:
Hey @Ruben2424, thanks for the tip. I'll make an issue for `set_priority` after this gets merged.

darioalessandro:
Hey @Ruben2424 what can we do to move this PR forward, seems like @seanmonstar is MIA

h3/src/config.rs
```diff
@@ -0,0 +1,43 @@
+use crate::proto::varint::VarInt;
+
+/// Configures the HTTP/3 connection
+#[derive(Debug, Clone, Copy)]
+pub struct Config {
```
seanmonstar, line 5: Is this type meant to be publicly exposed? If so, having all it's fields public will make it impossible to add any new options without it being a breaking change. This is why the builder pattern is used the way it is in `h2` and `hyper`.

h3/src/error.rs
```diff
@@ -37,12 +38,20 @@ impl PartialEq<u64> for Code {
     }
 }
 
+/// The error kind.
 #[derive(Clone)]
-pub(crate) struct ErrorImpl {
+pub struct ErrorImpl {
```
seanmonstar, line 43: I still feel it'd be best to keep this type private.

h3/src/ext.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+}
+
+/// HTTP datagram frames
+/// See: https://www.rfc-editor.org/rfc/rfc9297#section-2.1
+pub struct Datagram<B = Bytes> {
+    /// Stream id divided by 4
+    stream_id: StreamId,
+    /// The data contained in the datagram
+    pub payload: B,
```
seanmonstar, line 50: Should this field be private?

seanmonstar (commented):
I think the work you've done since I last reviewed is fantastic. It feels much more resilient to newer changes, and that "extending" is opt-in is excellent. Really great work! And thank you Ruben, your reviews helped accomplish that too!

I left some final thoughts inline, just a couple this time. They follow the same theme: protect the library from needing breaking changes, so we can optimize and change internals later.

h3/src/ext.rs
ten3roberts: It is to allow destructuring, but an `into_payload(self) -> B` and `payload(&self) -> &B` should suffice, but it doesn't allow partial destructuring

h3/src/config.rs
ten3roberts: I've marked the struct as non-excaustive for now.

It isn't really public either, as it is never exported `mod config`.

This is a pattern to allow using a type as a return value of a public function, without getting the "can not leak private type" While it is public, the module/symbol is never exported so the user can't mention it or store it in a struct, only use it with local variables given from a return value.

ten3roberts: I've made the fields public, and added getters for those need by webtransport

ten3roberts:
@seanmonstar I have now addressed your thoughts, could we run the CI once more?

Let me know what you think

PR #193, opened 2023-06-14
Fix `cargo doc` warning

chrislearn:


h3-webtransport/src/server.rs
```diff
@@ -34,7 +34,7 @@ use crate::stream::{BidiStream, RecvStream, SendStream};
 ///
 /// Maintains the session using the underlying HTTP/3 connection.
 ///
-/// Similar to [`crate::Connection`] it is generic over the QUIC implementation and Buffer.
+/// Similar to [`h3_quinn::Connection`](https://docs.rs/h3-quinn/latest/h3_quinn/struct.Connection.html) it is generic over the QUIC implementation and Buffer.
```
seanmonstar, line 37: I don't think it's supposed to be h3-quinn, since that's _not_ generic over the QUIC implementation, but uses quinn. Likely was something in `h3`, but may not be public.

chrislearn: Is https://docs.rs/h3/latest/h3/server/struct.Connection.html ?

seanmonstar: Oh, could be. Since this is server too.

seanmonstar (approved):

Issue #195, opened 2023-06-14
How to create customed WebtransportSession


chrislearn:
I want to create my customed WebtransportSession struct instead of using the WebtransportSession in h3_webtransport.

I found it is hard to reuse code in h3_webtransport for many struct has private new function. Is it possible public these new function?

Thanks

seanmonstar:
Is there something small we can open up that would let you use it? In general many of the internals are private so that we can change them in the future.

But we do want to be flexible where it is reasonable.

seanmonstar:
Closing for now. But if you or anyone else has requests for making things possible, you're welcome to ask! It's easier if we know a use case and can keep as much internals as possible private.

PR #196, opened 2023-06-15
For create customed WebtransportSession #195

chrislearn:


chrislearn:
Simply put, I want to reuse types other than `WebTransportSession` and create my own custom `WebTransportSession` to meet specific business requirements. 
For example, the current `WebTransportSession` is hard-coded with `Request<()>`, but in my specific case, I may have a custom Request object. 
I may even need to handle whether it is a `WEB_TRANSPORT` request and error handling outside of `WebTransportSession`.

h3/src/server.rs
```diff
@@ -820,6 +820,20 @@ pin_project! {
         }
 }
 
+impl<'a, C, B> ReadDatagram<'a, C, B>
+where
+    C: quic::Connection<B>,
+    B: Buf,
+{
+    #[allow(missing_docs)]
+    pub fn new(conn: &'a mut Connection<C, B>) -> Self {
```
Ruben2424, line 829: Thanks for the PR.
This should be parte of the feature `i-implement-a-third-party-backend-and-opt-into-breaking-changes` because it is not part of the http3 spec.

In general I am not sure if just adding `new` functions is the right way to go. 
The examples you mentions do not break the WebTransport RFC and sound reasonable to me. In my opinion this should be supported by our `WebTransportSession`. So the right step would be to improve our API.
What do you think?

chrislearn: `Request<()>` only be used in `fn validate_wt_connect(request: &Request<()>) -> bool`, is it possible move this validate outside and remove argument `request: Request<()>` from function accept?

Ruben2424: Would it help to make `WebTransport::accept(...)` generic like `WebTransport::accept<T>(request: Request<T>,...)`?

chrislearn: Unfortunately, I use `CustomRequest` here, not just `Request<Custom>`. 
Another issue is that even if `Request<()>` is needed, it seems that ownership is not required here, and `&Request<()>` will suffice.

chrislearn: Any new ideas or plans on this?

PR #199, opened 2023-06-17
move datagram to separate crate

Ruben2424:
The goal of This PR is to separate the [h3-datagram](https://datatracker.ietf.org/doc/html/rfc9297) logic from the `h3` crate providing a new crate `h3-datagram`.


h3-quinn/Cargo.toml
```diff
@@ -21,6 +21,7 @@ quinn = { version = "0.11", default-features = false, features = [
 tokio-util = { version = "0.7.9" }
 futures = { version = "0.3.28" }
 tokio = { version = "1", features = ["io-util"], default-features = false }
+h3-datagram = { path = "../h3-datagram" }
```
seanmonstar, line 24: Wondering, should it be optional? Is there enough code in here that people would benefit if they didn't need to compile it?

seanmonstar (commented):
Generally looks great, good work!

Might be good if @darioalessandro or @ten3roberts could look, to see if they can still use it.

h3-quinn/Cargo.toml
Ruben2424: > Wondering, should it be optional? Is there enough code in here that people would benefit if they didn't need to compile it?

Changed it.

Ruben2424:
> Generally looks great, good work!
> 
> Might be good if @darioalessandro or @ten3roberts could look, to see if they can still use it.

@seanmonstar i think you can continue with the review.
your comment got a :+1: . Additionally I tested against https://webtransport.day/ and it still worked.

ten3roberts:
I'm no longer using WT as much, but I had a look and it looks great.

Nothing suspicious in the code, and good use of pub/pub(crate) and doc-hidden where needed :+1:

seanmonstar (approved):

PR #200, opened 2023-06-25
simplify traits 

Ruben2424:
This is a not 100% working change to explore some of the proposed changes from #177.


h3/src/quic.rs
```diff
@@ -48,21 +62,11 @@ pub trait Connection<B: Buf> {
...
-    fn poll_accept_recv(
-        &mut self,
-        cx: &mut task::Context<'_>,
-    ) -> Poll<Result<Option<Self::RecvStream>, Self::Error>>;
-
-    /// Accept an incoming bidirectional stream
-    ///
-    /// Returning `None` implies the connection is closing or closed.
-    fn poll_accept_bidi(
+    /// Poll the connection for incoming streams.
```
FlorianUekermann, line 65: Add: `/// Returning an Err implies the connection is closing or closed.`

FlorianUekermann (commented):
The poll merger looks good to me. Since we need to document expectations for different close scenarios anyway, it may make sense to handle the close type situation in this PR as well.  I would recommend the enum proposal from #177 B2 (replace Error with Close enum).

In any case, I think the expectation that the quic implementation always emits a `Close`/`Error` via `poll_incoming` on termination should be documented.

Stream opening itself can't fail (only block), unless the connection is terminated. Therefore I think it would be good to remove the errors from `poll_open_*`. That removes any ambiguity how Closes are communicated (only via `poll_incoming`). I'm not even convinced returning an `Option` is necessary. Just Poll::Pending should be fine as well.

Ruben2424:
Thanks for the review. 

>I would recommend the enum proposal from https://github.com/hyperium/h3/issues/177 B2 (replace Error with Close enum).

I will continue to work on this and implement #177 B2.

I Think this are my next tasks:

- [ ] Fix the failing tests
- [ ] Implement #177 B2
- [ ] Make sure everything is well Documented
- [ ] Make WebTransport work with all the changes

dongcarl:
@Ruben2424 Is this PR in a state where it's ready for review?

Ruben2424:
> @Ruben2424 Is this PR in a state where it's ready for review?

@dongcarl you can review the changes if you want. But there is still some work left to improve the error types.
I probably have time to continue on this next week.

Ruben2424:
I think i will not continue on this as one big pr. Instead I probably will split them up into multiple smaller ones which are easier to review.

PR #202, opened 2023-07-11
Bump `H3_DATAGRAM` setting ID according to RFC9297

dongcarl:


Ruben2424 (approved):
Nice catch. Thank You!

PR #203, opened 2023-07-11
Initial support for RFC 9298 "Proxying UDP in HTTP"

dongcarl:
1.  Add support for RFC 9298 in `:protocol` extension
2. bugfix: Properly encode `:protocol` header
    ```
    Previously, the `:protocol` header wouldn't be set when the `Protocol`
    extension was active for a `Request`.
    
    This was most likely because the Webtransport PR (the PR that introduced
    the `Protocol` extension) did not include a client implementation and
    thus did not test this.
    ```

### DIFF
--- a/h3/src/ext.rs
+++ b/h3/src/ext.rs
@@ -22,6 +22,15 @@ impl Protocol {
     pub const WEB_TRANSPORT: Protocol = Protocol(ProtocolInner::WebTransport);
     /// RFC 9298 protocol
     pub const CONNECT_UDP: Protocol = Protocol(ProtocolInner::ConnectUdp);
+
+    /// Return a &str representation of the `:protocol` pseudo-header value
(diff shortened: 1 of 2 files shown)


dongcarl:
@ten3roberts You might be interested in reviewing this 😬

ten3roberts:
Looks interesting, though I am unfortunately not a collaborator on this repository so I don't think I can _submit_ a review.

But yes, you are indeed correct, the missing serialization of the protocol extension header is just an oversight as the Webtransport PR was only for the server due to PR size and usage requests.

Ruben2424 (approved):
looks good to me, thank you!

@ten3roberts We appreciate the value of feedback and reviews from everyone, whether you are a collaborator or not. You are welcome to submit a review if you would like to share your thoughts.

PR #204, opened 2023-07-12
bugfix: Actually encode extensions in header

dongcarl:
This was previously attemped in 0dffc9008f65dc0aea272623a5f17c38df37f737, but I didn't expect that all of the request's extensions have been ignored when passed to `Header::request`, this fixed it.

A previous attempt (not mine) is found here: #159

### DIFF
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -159,9 +159,10 @@ where
             method,
             uri,
             headers,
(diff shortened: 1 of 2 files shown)


Ruben2424 (approved):
Oh, i missed that too. Thanks.

PR #207, opened 2023-07-17
Add a test-only `send_settings` config option

dongcarl:
1. tests: Remove incorrect driver comments
    ```
    In all of these test cases, the driver _is_ polled.
    ```
1. tests: Poll driver in header_too_big_discard_from_client
    ```
    It is actually safe to poll the driver, since:
    
    1. It only matters that the server has a false impression of the
       client's max_field_section_size and
    2. It doesn't matter if the client knows about the server's
       max_field_section_size or not
    ```
1. Add a test-only send_settings config option
    ```
    Instead of reading into the server's RequestStream and manually
    overriding its shared state, just have a test-only config option to have
    the client not send its settings.
    ```

### DIFF
--- a/h3/src/client.rs
+++ b/h3/src/client.rs
@@ -494,6 +494,12 @@ impl Builder {
         }
     }
 
+    #[cfg(test)]
+    pub fn send_settings(&mut self, value: bool) -> &mut Self {
+        self.config.send_settings = value;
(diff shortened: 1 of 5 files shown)


Ruben2424 (approved):
Good job. Thank you!

PR #208, opened 2023-07-17
Split out a `Settings` struct from `Config`

dongcarl:
The `Settings` struct is an abstract representation of `frame::Settings`.

Basically:

- When talking about our own connection, we should use `Config`.
  - e.g. in `ConnectionInner`'s `config` field

- When talking about a peer's settings, we should use `Settings.
  - e.g. in `SharedState`'s `peer_config` field

This separation allows us to define some conversions:

- `Config` -> `frame::Settings`
  - used in `ConnectionInner::new`

- `&frame::Settings` -> `Settings`
  - used in `ConnectionInner::poll_control`

### DIFF
--- a/h3/src/config.rs
+++ b/h3/src/config.rs
@@ -1,4 +1,9 @@
-use crate::proto::varint::VarInt;
+use std::{
+    convert::TryFrom,
+    ops::{Deref, DerefMut},
+};
+
+use crate::proto::{frame, varint::VarInt};
 
 /// Configures the HTTP/3 connection
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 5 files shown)


Ruben2424 (commented):
Looks good to me. Thanks.
It seams that there are some conflict changes with your previous PR. Would you mind resolving them?

dongcarl:
@Ruben2424 Resolved!

dongcarl:
Perhaps one thing to discuss is whether or not to include the `Deref` and `DerefMut` impls here: https://github.com/hyperium/h3/pull/208/files#diff-4fc440b791f25fbde8061757f9ecfdcb61af28bc20a5b3c236f0ae1f6c96db4eR181-R194

https://doc.rust-lang.org/std/ops/trait.Deref.html says that: "Deref should only be implemented for smart pointers"

h3/src/config.rs
```diff
@@ -58,14 +170,25 @@ impl Config {
...
-            enable_webtransport: false,
-            enable_extended_connect: false,
-            enable_datagram: false,
-            max_webtransport_sessions: 0,
+            settings: Default::default(),
         }
     }
 }
+
+impl Deref for Config {
```
Ruben2424, line 181: > Perhaps one thing to discuss is whether or not to include the Deref and DerefMut impls here: https://github.com/hyperium/h3/pull/208/files#diff-4fc440b791f25fbde8061757f9ecfdcb61af28bc20a5b3c236f0ae1f6c96db4eR181-R194
> https://doc.rust-lang.org/std/ops/trait.Deref.html says that: "Deref should only be implemented for smart pointers"

`Config` is not a Smart Pointer so I think we should remove them to avoid confusion.

dongcarl:
@Ruben2424 Removed!

Ruben2424 (approved):
Looks very good. Thanks!

PR #210, opened 2023-10-06
fix: authority check

yaotthaha:


Ruben2424:
Hi, Thanks for the PR.

As I understand the [specification](https://www.rfc-editor.org/rfc/rfc9114#name-request-pseudo-header-field) , the current implementation is correct. Because the [host](https://docs.rs/http/0.2.9/http/uri/struct.Authority.html#method.host) method only returns the `host` without the port or userinfo.


> To ensure that the HTTP/1.1 request line can be reproduced accurately, this pseudo-header field MUST be omitted when translating from an HTTP/1.1 request that has a request target in a method-specific form; see [Section 7.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.1) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9114#RFC9110)]. Clients that generate HTTP/3 requests directly SHOULD use the :authority pseudo-header field instead of the Host header field. An intermediary that converts an HTTP/3 request to HTTP/1.1 MUST create a Host field if one is not present in a request by copying the value of the :authority pseudo-header field.

> If the :scheme pseudo-header field identifies a scheme that has a mandatory authority component (including "http" and "https"), the request MUST contain either an :authority pseudo-header field or a Host header field. If these fields are present, they MUST NOT be empty. If both fields are present, they MUST contain the same value. If the scheme does not have a mandatory authority component and none is provided in the request target, the request MUST NOT contain the :authority pseudo-header or Host header fields.


Do you have a specific problem leading to this PR?

PR #211, opened 2023-10-22
increase msrv and update dependencies 

Ruben2424:


h3-quinn/Cargo.toml
```diff
@@ -19,6 +19,6 @@ quinn = { version = "0.10", default-features = false, features = [
     "futures-io",
 ] }
 quinn-proto = { version = "0.10", default-features = false }
-tokio-util = { version = "0.7.7" }
-futures = { version = "0.3.27" }
-tokio = { version = "1.28", features = ["io-util"], default-features = false }
+tokio-util = { version = "0.7.9" }
+futures = { version = "0.3.28" }
+tokio = { version = "1.33", features = ["io-util"], default-features = false }
```
seanmonstar, line 24: Is such a newer version of Tokio required? If not, we can keep the lower number, so it works with older versions. Anyone who had a newer one still works too.

Ruben2424: Not required i think. I will change that.

h3-quinn/Cargo.toml
```diff
@@ -19,6 +19,6 @@ quinn = { version = "0.10", default-features = false, features = [
     "futures-io",
 ] }
 quinn-proto = { version = "0.10", default-features = false }
-tokio-util = { version = "0.7.7" }
-futures = { version = "0.3.27" }
-tokio = { version = "1.28", features = ["io-util"], default-features = false }
+tokio-util = { version = "0.7.9" }
+futures = { version = "0.3.28" }
+tokio = { version = "1", features = ["io-util"], default-features = false }
```
seanmonstar, line 24: Is it true we work on 1.0? Usually we keep the minimum version that works. I suspect we use something that was newer than that. It's possible to check with `minimal-versions`.

Ruben2424: Hmm `cargo minimal-versions check --workspace` is successful on my machine. Is there something else to check?

seanmonstar: Alrighty, sounds good!

seanmonstar (approved):

PR #214, opened 2023-10-28
ensure LICENSE file is present in all published crates

decathorpe:
This PR adds symbolic links for the `LICENSE` file into the `h3` and `h3-quinn` workspace members so that `cargo package` / `cargo publish` picks them up and includes them when publishing crates.

The MIT license requires that (re)distributed sources contain a copy of the license text. Not including the license text is a problem, especially for Linux distributions that package this crate as a dependency, for example, for the HTTP/3 support in the `reqwest` crate.

I have not added a symlink into the `h3-webtransport` directory since this workspace member is not published to crates.io, and does even not have complete cargo metadata yet.


Ruben2424 (approved):
Thanks!

PR #220, opened 2023-12-08
expose poller varieties of recv_data and send_data

gopakumarce:
It is useful in certain applications to have poller apis exposed for equivalent of recv_data and send_data.

recv_data is being mapped to poll_read in this diff send_data is a combination of push_data and poll_ready

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -48,5 +48,13 @@ name = "server"
 path = "server.rs"
 
 [[example]]
+name = "client_streaming"
+path = "client_streaming.rs"
+
+[[example]]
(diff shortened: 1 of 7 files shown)


gopakumarce:
@seanmonstar - pls take a gander at whenever you get a chance, ive also added a client_streaming.rs and server_streaming.rs to the examples/

gopakumarce:
Closing this, even this will not help with a poll-mode send and recv for h3. Basically h3 will need to send and recv in an async task and then siphon off the data into some channel which can be polled, thats the only model an async read/write can be implemented on top of an h3 stream

PR #228, opened 2024-02-12
add 'Connection:poll_recv_data()'

SergioBenitez:
As in the data. Needed to call from a `Stream` impl.


SergioBenitez:
Did not see https://github.com/hyperium/h3/pull/222 before. This is basically a cleaner implementation of that.

seanmonstar (approved):

PR #229, opened 2024-03-14
split `server.rs` and `client.rs` into multiple files in thair own module

Ruben2424:
I split the big files `server.rs` and `client.rs` into multiple files and introduced the `server` and `client` modules to gain a clearer project structure.

Also fixed some (clippy) warnings.


examples/client.rs
```diff
@@ -113,7 +113,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // h3_quinn implements the trait w/ quinn to make it work with h3.
     let quinn_conn = h3_quinn::Connection::new(conn);
 
-    let (mut driver, mut send_request) = h3::client::new(quinn_conn).await?;
+    let (mut driver, mut send_request) = h3::client::builder::new(quinn_conn).await?;
```
seanmonstar, line 116: Should we keep some of these common functions/types at the `h3::client` and `h3::server` level, re-exported from the submodules? That way, we can have more structure internally, and users don't need to care how we do it.

seanmonstar (approved):
Looks like a nice clean up to me!

PR #231, opened 2024-03-20
Ensure field names do not contain uppercase

Pi-Cla:
I classify this as implication because the http crate is responsible for ensuring HeaderName::from_lowercase throws an error when it encounters uppercase characters


seanmonstar (approved):
Thank you (and for remembering to update the compliance)!

PR #232, opened 2024-03-20
Ensure authority and/or host fields are not empty

Pi-Cla:
We could have an authority or host that consists of an empty string


Ruben2424:
Hi, thanks for the PR!
You are using the `try_from` method in the test to create the header. This method is also used wen receiving a request.
`try_from` parses the header field to a typed header field from the `http` crate which allows no empty values.

I would leave the check and the test for the `host` field where it is.
For the `authority` field, i would copy the compliance comment [here](https://github.com/hyperium/h3/blob/b44edeb60d234d49c45828395108f7519a048d4b/h3/src/proto/headers.rs#L294) and assert the result of `try_from` instead of `into_request_parts`

h3/src/proto/headers.rs
```diff
@@ -202,6 +201,9 @@ impl Iterator for HeaderIter {
     }
 }
 
+//= https://www.rfc-editor.org/rfc/rfc9114#section-4.3.1
```
Ruben2424, line 204: Looks good so far. Do you mind moving that comment to the `Field::parse` method? The parse method will return the error when `authority` is empty. So one can find the location when looking through the compliance report.
iirc the error comes from here:
https://github.com/hyperium/h3/blob/b44edeb60d234d49c45828395108f7519a048d4b/h3/src/proto/headers.rs#L294

Ruben2424 (approved):

PR #233, opened 2024-03-21
Convert field names to lower before encoding

Pi-Cla:
I feel like there is probably a more elegant solution... so if anyone has suggestions, throw em in here


h3/src/client/connection.rs
```diff
@@ -142,7 +144,19 @@ where
...
-        let headers = Header::request(method, uri, headers, extensions)?;
+
+        //= https://www.rfc-editor.org/rfc/rfc9114#section-4.2
+        //# Characters in field names MUST be
+        //# converted to lowercase prior to their encoding.
+        let mut lower_headers = HeaderMap::new();
+
+        for (n, v) in headers.iter() {
+            lower_headers.append(
+                HeaderName::from_lowercase(n.as_str().to_ascii_lowercase().as_bytes()).unwrap(),
```
seanmonstar, line 155: All `HeaderName`s store themselves as lowercase. You don't need to convert them here. The `from_lowercase` only makes sure the bytes were lowercase originally, vs `from_bytes` will allow uppercase but convert it.

PR #234, opened 2024-03-22
do not explicit copy data before sending

Ruben2424:
I am not sure if this is the best way to write this or if it gains performance. I am happy to get Feedback.

### DIFF
--- a/h3-quinn/src/lib.rs
+++ b/h3-quinn/src/lib.rs
@@ -590,11 +590,11 @@ where
             match res {
                 Ok(cnt) => data2.advance(cnt),
                 Err(err) => {
+                    self.writing = Some(data2);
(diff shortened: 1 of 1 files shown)


seanmonstar:
Good idea! Do things have to be `Send`? If we don't need it, people can choose to use single threaded types.

Ruben2424:
> Good idea! Do things have to be `Send`? If we don't need it, people can choose to use single threaded types.

@seanmonstar I think they have to be `Send` because `data` is moved into the `ReusableBoxFuture`.
Maybe i could bring back the old implementation where `B: !Send`?

Ruben2424:
>  Maybe i could bring back the old implementation `where B: !Send`?
Ok i think this is not possible.

Ruben2424:
fixed in #279

PR #236, opened 2024-04-13
Add WebSocket :protocol extension per RFC 9220.

vi:
Attempt to add WebSocket support to h3.

I don't know any server that supports RFC 9220, so it is untested.


pythonwood:
Hi. you can add some unit test.
support rfc9220 server:
https://github.com/ngtcp2/nghttp3

vi:
What should the unit test test? `as_str` / `from_str` implementations?

ararog:
Based on what I see on other implementations of websocket, h3 should only provide the minimum at http level to be able to establish websockets connection, which is, in short, the CONNECT pseudo-headers, and the settings frames, besides that, looks like the remaining parts can still take advantage of existing frame parsing libraries.

vi:
Refreshed the pull request after a conflict from #273, which has a very similar idea.

PR #237, opened 2024-05-03
Clean up warnings

djc:


h3/src/stream.rs
```diff
@@ -250,7 +250,7 @@ where
     B: Buf,
 {
     Control(FrameStream<S, B>),
-    Push(u64, FrameStream<S, B>),
```
Ruben2424, line 253: should this warning be suppress? I think this field is needed for Server Push (which is not implemented yet).

djc: Can just add the field back when the feature is implemented, IMO.

Ruben2424: i don't have a strong opinion about this. So ok for me.

Ruben2424 (approved):
lgtm

PR #238, opened 2024-05-03
Upgrade to Quinn 0.11

djc:
This is still failing some tests -- would appreciate some help with addressing these.

`control_close_send_error` fails for me on master, too.

We plan to release 0.11 soon, see https://github.com/quinn-rs/quinn/issues/1737.

### DIFF
--- a/h3/Cargo.toml
+++ b/h3/Cargo.toml
@@ -35,14 +35,14 @@ fastrand = "2.0.1"
 assert_matches = "1.5.0"
 futures-util = { version = "0.3", default-features = false, features = ["io"] }
 proptest = "1"
-quinn = { version = "0.10", default-features = false, features = [
+quinn = { version = "0.11", default-features = false, features = [
     "runtime-tokio",
-    "tls-rustls",
+    "rustls",
     "ring",
 ] }
-quinn-proto = { version = "0.10", default-features = false }
-rcgen = "0.12"
-rustls = "0.21"
+quinn-proto = { version = "0.11", default-features = false }
+rcgen = "0.13"
+rustls = { version = "0.23", default-features = false, features = ["logging", "ring", "std"] }
 tokio = { version = "1", features = ["rt", "macros", "io-util", "io-std"] }
 tracing-subscriber = { version = "0.3", default-features = false, features = [
     "fmt",
--- a/h3/src/tests/request.rs
+++ b/h3/src/tests/request.rs
@@ -1434,7 +1434,7 @@ where
         let mut buf = BytesMut::new();
         request(&mut buf);
         req_send.write_all(&buf[..]).await.unwrap();
-        req_send.finish().await.unwrap();
+        req_send.finish().unwrap();
 
         let res = req_recv
             .read(&mut buf)
--- a/h3/src/tests/request.rs
+++ b/h3/src/tests/request.rs
@@ -69,6 +69,8 @@ async fn get() {
             .await
             .expect("send_data");
         request_stream.finish().await.expect("finish");
+
+        let _ = incoming_req.accept().await.unwrap();
     };
 
     tokio::join!(server_fut, client_fut);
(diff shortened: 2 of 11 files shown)


Ruben2424:
i probably can take a look at the tests this weekend

inflation:
The same error used to happen in particularly macOS a lot.

djc:
Yeah, I was testing on macOS.

Cargo.toml
```diff
@@ -8,3 +8,7 @@ members = [
     "examples",
 ]
 resolver = "2"
+
+[patch."crates-io"]
+quinn = { git = "https://github.com/quinn-rs/quinn", branch = "no-default-features" }
```
Ruben2424, line 13: @djc i think the branch `no-default-features` is not available anymore.
Should i test with `main` or do you have a specific branch?

djc: Main is good -- the branch has been merged. Thanks for having a look!

Ruben2424:
I fixed some race conditions in the failing tests. Does `control_close_send_error` now work on macOS?
The test `tests::connection::client_close_only_on_last_sender_drop` still fails. I couldn't fix it without understanding what exactly it tests.
Does it test anything? Can we remove it?

djc:
Yes, I only have `client_close_only_on_last_sender_drop` failing now.

djc:
> The test `tests::connection::client_close_only_on_last_sender_drop` still fails. I couldn't fix it without understanding what exactly it tests.
> Does it test anything? Can we remove it?

@seanmonstar are you the person who has more context on this?

h3/Cargo.toml
```diff
@@ -35,14 +35,14 @@ fastrand = "2.0.1"
...
-    "tls-rustls",
+    "rustls",
     "ring",
 ] }
-quinn-proto = { version = "0.10", default-features = false }
-rcgen = "0.12"
-rustls = "0.21"
+quinn-proto = { version = "0.11", default-features = false }
+rcgen = "0.13"
+rustls = { version = "0.23", default-features = false, features = ["logging", "ring", "std"] }
```
zh-jq, line 45: How about disable the `logging` feature by default?

Ruben2424: this are just dev-dependencies for unittests. h3 does not depend on rustls directly.

Ruben2424:
> One option is to comment it out with a note, and if we notice something wrong related to closure, then we can try to figure out what this one was checking specifically.

I think the failure has nothing todo with the purpose of this test. It is also a race condition. The client future finishes so fast, that the server cannot initialize the connection fully before the client closes the connection.
Then this `unwrap()` panics `let mut incoming = server::Connection::new(conn).await.unwrap();`.

We could also do a `recv_response().await` on the client request streams. But for me it seams that maybe they were intentionally omitted? idk?

seanmonstar:
I don't have anything else. You could comment out the test, or explore what Ruben said.

Ruben2424:
I can try to finnish the test (or comment it out) this weekend. After a review i can prepare releases.

djc:
@Ruben2424 thanks!

Ruben2424:
The tests are now successful. @seanmonstar do you mind doing a review?

h3/src/tests/request.rs
```diff
@@ -398,20 +419,19 @@ async fn header_too_big_client_error() {
...
             );
         };
-        tokio::select! {biased; _ = req_fut => (),_ = drive_fut => () }
+        tokio::join! {req_fut, drive_fut }
     };
 
     let server_fut = async {
         let conn = server.next().await;
-        //= https://www.rfc-editor.org/rfc/rfc9114#section-4.2.2
-        //= type=test
```
seanmonstar, line 407: Why remove this comment? It's used for RFC coverage.

seanmonstar (approved):

h3/src/tests/request.rs
Ruben2424: if I am correct this does not test the server limit. It tests that our client does not send any requests that extend the server limit.

djc:
Thanks for the reviews/fixes, looking forward to the release!

PR #242, opened 2024-06-03
make h3 able to connect to nginx

Ruben2424:
This starts the qpack streams, but does not use them.
Nginx ends the Connection with a Connection error when receiving h3s grease stream when they are not present.

The creation of `ConnectionInner` does now not wait for the grease stream to finnish. The grease stream is now handled in the background in the `poll_control`.


seanmonstar:
Is there a public endpoint somewhere using nginx h3? So I can try it out with master and with this PR?

Ruben2424:
> Is there a public endpoint somewhere using nginx h3? So I can try it out with master and with this PR?

I think https://quic.nginx.org/ . But I tested locally.

seanmonstar (approved):

PR #243, opened 2024-06-15
remove unused waker

Ruben2424:
if I am not completely wrong, this `waker` does nothing, right?


Ruben2424:
I am not sure if I understand `wakers` completely. But as far as i understand this only wakes the current task. So when the Connection is moved to a different task, `Connection` is not waked right? Would it be better to replace it with a oneshot channel instead of removing it?

PR #247, opened 2024-06-21
add tracing instrumentation

Ruben2424:
This adds tracing to most connection relevant functions. This should make it easier to debug.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -111,6 +111,11 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         toolchain: [stable, beta]
+        include:
+          - features: i-implement-a-third-party-backend-and-opt-into-breaking-changes
+          - features: enable-tracing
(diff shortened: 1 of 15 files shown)


seanmonstar:
It'd be nice if tracing could be optional. I was recently reminded of the desire for this in `h2`: https://github.com/hyperium/h2/issues/713

Ruben2424:
@seanmonstar, I added features for h3 and h3-quinn to enable tracing. What do you think?

Ruben2424:
> Looks good to me. For what it's worth, you can name it `tracing`. Under features in the Cargo.toml, it could be `tracing = ["dep:tracing"]`. I think something like that is cleaner, it doesn't need to include the prefix of the crate (like `h3-`) in its own features.

I added the prefix because it confused me to have two crates with the same feature in one repo. When I call `cargo test --all --features=tracing` does it compile h3 or h3-quinn with the tracing feature. Or both?

PR #255, opened 2024-07-08
add send datagram support in h3 client

needon1997:


Ruben2424:
Thanks for the PR. In my opinion this code (also server side) belongs in a different crate as it is not part of the http3 spec.

I started moving things in #199 but dadn't time to finnish. If you want, feel free to continue the PR.

PR #258, opened 2024-08-02
increase msrv to 1.70 like tokio

Ruben2424:


djc (approved):
Makes sense!

seanmonstar:
Do we need anything from the newer Tokio? If not, we could just pin the Tokio version in the MSRV job. That's what I've done in hyper, h2, and reqwest.

djc:
> Do we need anything from the newer Tokio? If not, we could just pin the Tokio version in the MSRV job. That's what I've done in hyper, h2, and reqwest.

For Quinn we've chosen (IIRC) to just adopt the Tokio MSRV, so that might constrain the choices here?

Ruben2424:
Quinn has 1.70 msrv, so I guess h3-quinn will have to use the same.
if we want to support earlier versions for h3 we can change it back. But I would not complicate thinks now. 1.70 is over a year old and h3 is still in pre release.

PR #263, opened 2024-12-03
Avoiding extra allocation for shared error.

nurmohammed840:


seanmonstar:
Thanks! I started #264 to fix up that CI failure.

PR #267, opened 2024-12-11
Added .id() for client RequestStream

yaroslavros:
Currently only server RequestStream has ability to get StreamId. However would also be a very useful capability on the client side - especially for synchronisation of streams with datagrams.


Ruben2424 (approved):
lgtm, thanks

PR #269, opened 2024-12-17
Expose poll_recv_trailers APIs

youyuanwu:
Expose poll style APIs for receive trailers (and data).
I need these APIs to imbed the h3 connection into http body and implement the poll_frame. With this change I can successfully send and receive http::Request body with streaming.

This also matches the h2 poll_trailers api:
https://github.com/hyperium/h2/blob/3bce93e9b0dad742ffbf62d25f9607ef7651a70c/src/share.rs#L427

### DIFF
--- a/h3/src/client/stream.rs
+++ b/h3/src/client/stream.rs
@@ -12,7 +12,10 @@ use crate::{
     qpack,
     quic::{self},
 };
-use std::convert::TryFrom;
+use std::{
+    convert::TryFrom,
+    task::{Context, Poll},
+};
 
 /// Manage request bodies transfer, response and trailers.
 ///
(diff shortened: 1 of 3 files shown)


h3/src/connection.rs
```diff
@@ -826,20 +828,18 @@ where
...
+                Some(_) => return Poll::Ready(Err(Code::H3_FRAME_UNEXPECTED.into())),
+                None => return Poll::Ready(Ok(None)),
             }
         };
-
         if !self.stream.is_eos() {
             // Get the trailing frame
-            let trailing_frame = future::poll_fn(|cx| self.stream.poll_next(cx))
-                .await
+            let trailing_frame = futures_util::ready!(self.stream.poll_next(cx))
```
Ruben2424, line 837: I think this introduces a bug. There is no error returned for the case that 2 `Headers` frames are sent by the peer and the first call of `poll_next`  returns `Poll::Pending`.

h3/src/client/stream.rs
```diff
@@ -162,6 +173,20 @@ where
         res
     }
 
+    /// Poll receive an optional set of trailers for the response.
+    pub fn poll_recv_trailers(
+        &mut self,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Option<HeaderMap>, Error>> {
+        let res = self.inner.poll_recv_trailers(cx);
+        if let Poll::Ready(Err(e)) = &res {
```
Ruben2424, line 182: this code is redundant to the async `recv_trailers` version. I think it would be better to remove the `
    pub async fn recv_trailers(&mut self) -> Result<Option<HeaderMap>, Error> {` in `h3/src/connection.rs` and call this function within the `pub async fn recv_trailers` function. And the same on the server side.

This reduces the redundant error handling.

Ruben2424 (changes requested):
Thanks for the PR. I left some thoughts below.

h3/src/client/stream.rs
youyuanwu: Done.

h3/src/connection.rs
youyuanwu: Fixed by saving the trailers when pending is encountered for the trailing_frame.
The first header frame is always used for parsing the trailers. If peer sends more header frames, we poll it until it is received and checked. This assumes is_eos() is still false after poll_next returned pending.

Ruben2424 (approved):
lgtm

PR #273, opened 2025-01-14
Added connect-ip (RFC9484) as a known HTTP/3 CONNECT protocol

yaroslavros:
Fairly trivial update to recognize connect-ip as a valid :protocol.


seanmonstar (approved):
I think adding this is fine.

As long as it doesn't require `h3` to _do_ something different, and instead the user can implement everything outside of the library, great!

PR #274, opened 2025-01-14
Client also needs an ability to stop streams with error code

yaroslavros:
For some reason stop_stream is currently only available for server streams.

However, it is also useful for clients to be able to stop streams with an error code - for example on disconnect of CONNECT socket or to close stream on receiving malformed capsule when using capsule protocol inside h3 stream.


Ruben2424 (approved):
lgtm

PR #278, opened 2025-01-22
ci: add .duvet/config.toml

camshaft:
This change updates the compliance tool `duvet` to version 0.4.0. This update includes

* A `.duvet/config.toml`, which makes it easier to define the workspace sources and specifications, rather than passing it all on the command line
* A `snapshot` report, which enables the project to avoid accidental regressions in requirement coverage. You can read more about it here: https://awslabs.github.io/duvet/reports.html#snapshot

I've also moved all of the `specs` files to the `.duvet` directory, which is now the preferred location to store everything related to `duvet` .


ci/compliance/report.sh
```diff
@@ -2,16 +2,6 @@
...
-    --spec-pattern 'ci/compliance/specs/**/*.toml' \
-    --source-pattern 'h3/**/*.rs' \
-    --workspace \
-    --exclude duvet \
-    --require-tests false \
-    --blob-link "https://github.com/hyperium/h3/blob/master" \
-    --issue-link 'https://github.com/hyperium/h3/issues' \
-    --no-cargo \
-    --html ci/compliance/report.html
+duvet report
```
Ruben2424, line 5: Do we need the --ci parameter?

Ruben2424 (commented):
Just to clarify for me. There is no mechanism to keep `snapshot.txt` up to date in the CI right? Or did I miss it?
So when new citations are added it is required to call `duvet report` locally and push the updated  `snapshot.txt` or how does it work?

camshaft:
> So when new citations are added it is required to call duvet report locally and push the updated snapshot.txt or how does it work?

That's exactly right. As you change the statuses of requirements locally, you check in the updates to `snapshot.txt`. This acts as an explicit acknowledgement that the snapshot changed and you commit it with the change that altered it. Then in CI, `duvet report` will enforce that the derived `snapshot.txt` matches what you have committed to the repository.

This is the same kind of mechanism that something like https://insta.rs/ provides.

ci/compliance/report.sh
camshaft: duvet should now look at the the `CI` environment variable, which is set by [GitHub actions by default](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables). The nice thing about using the env var is the same script can be used locally to update the snapshot instead of checking it.

Ruben2424 (approved):

PR #280, opened 2025-02-25
Add datagram support for client

pinkisemils:
As per #253 , these changes add the ability for a client to send HTTP datagrams over the underlying QUIC connection. I am not entirely sure about how the underlying QUIC connection is being exposed - I am open to better alternatives :)

PR #282, opened 2025-03-10
Add extended CONNECT setting for client conn

shitz:
Various HTTP/3 extensions that use HTTP datagrams such as [connect-udp](https://datatracker.ietf.org/doc/html/rfc9298), [connect-ip](https://www.rfc-editor.org/rfc/rfc9484.html), or [webtransport](https://datatracker.ietf.org/doc/draft-ietf-webtrans-overview/) rely on extended CONNECT. #280 added datagram support for client connections, however, to use it for any of the aforementioned protocols, we also need to be able to signal extended CONNECT support in the H3_SETTINGS.

### DIFF
--- a/h3/src/client/builder.rs
+++ b/h3/src/client/builder.rs
@@ -104,6 +104,12 @@ impl Builder {
         self
     }
 
+    /// Enables the extended CONNECT protocol required for various HTTP/3 extensions.
+    pub fn enable_connect(&mut self, value: bool) -> &mut Self {
+        self.config.settings.enable_extended_connect = value;
+        self
+    }
+
     /// Create a new HTTP/3 client from a `quic` connection
     pub async fn build<C, O, B>(
         &mut self,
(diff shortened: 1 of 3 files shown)


h3/src/client/builder.rs
```diff
@@ -104,6 +104,12 @@ impl Builder {
         self
     }
 
+    /// Enables the extended CONNECT protocol required for various HTTP/3 extensions.
+    pub fn enable_connect(&mut self, value: bool) -> &mut Self {
```
skreborn, line 108: Shouldn't the method be called `enable_extended_connect`?

seanmonstar: There's no API stability at the moment while at 0.0.x. All upgrades are manual.

shitz: Renamed in `client::builder()` and `server::builder()`.

Ruben2424 (approved):
lgtm

PR #290, opened 2025-03-18
h3-quinn: fix usage of a private StreamId field

seanmonstar:
We were accessing a field that was `doc(hidden)`. A new version of quinn-proto made this field actually private, but made it possible to convert a stream ID into a `u64`.


djc (approved):
LGTM. Might want to include a version bump?

seanmonstar:
I do version bumps as part of a release commit.

Issue #284, opened 2025-03-13
test failure on 32-bit.


plugwash:
After updating the h3 crate in Debian I discovered that tests fail on 32-bit systems.

    ---- qpack::prefix_int::test::codec_5_bits stdout ----
    thread 'qpack::prefix_int::test::codec_5_bits' panicked at src/qpack/prefix_int.rs:101:67:
    called `Result::unwrap()` on an `Err` value: Overflow
    
    ---- qpack::prefix_int::test::codec_8_bits stdout ----
    thread 'qpack::prefix_int::test::codec_8_bits' panicked at src/qpack/prefix_int.rs:101:67:
    called `Result::unwrap()` on an `Err` value: Overflow

Digging into the git history I found that the encode and decode functions previously used usize and were changed to use u64 instead of usize ( https://github.com/hyperium/h3/pull/223/ ) however the value of MAX_POWER was not changed and as a result decoding large numbers still fails.

I believe the correct fix is to use the same MAX_POWER value on all targets and intend to implement that in Debian, but feedback would be appreciated.

seanmonstar:
Your fix sounds right to me. Want to submit a PR?

plugwash:
Patch is at https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/h3/debian/patches/allow-decoding-large-numbers-on-32-bit.patch

PR #301, opened 2025-05-03
Integer overflow when parsing prefixed integers from qpack

Ruben2424:
Changes in this PR:
- solves #284 by applying patch from https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/h3/debian/patches/allow-decoding-large-numbers-on-32-bit.patch
- Adds 32bit CI so tests run also on 32bit
- Allows only 9 instead of 10 follow up bytes in a qpack prefix integer to prevent integer overflow

# Overflow
previous to this PR the prefix_int `MAX_POWER` restricted to 10 * 7 (70) bits, which does not fit in an u64.  
This PR restricts the `MAX_POWER` to 9 * 7 (63) bits. This still meets the requirement from the qpack [spec](https://www.rfc-editor.org/rfc/rfc9204.html#name-prefixed-integers)  of 62 bit:

> QPACK implementations MUST be able to decode integers up to and including 62 bits long.

This means our implementation now rejects some values that would fit in `u64`.   
Since the spec only requires 62 bit i chose this approach instead of adding more logic to the decoder.


seanmonstar (approved):

Ruben2424:
@seanmonstar had to change a few things to satisfy CI after your review. Some unrelated to the original change:

I moved the `server_and_client_should_connect_successfully` rust test which spawned processes to a script which the ci now runs directly. Idk what exactly caused this to fail, but now it works.

PR #317, opened 2025-08-28
feat(h3-quinn): update quinn to 0.11.9

CosminPerRam:
I'd suggest to explicitly update `quinn` to `0.11.9` to use the new version of `socket2` (`0.6`), many other projects have already migrated to it and personally would want to do so to reduce duplicated dependencies.

Hope this is an appropiate request, thanks.

### DIFF
--- a/h3-quinn/Cargo.toml
+++ b/h3-quinn/Cargo.toml
@@ -15,7 +15,7 @@ license = "MIT"
 [dependencies]
 h3 = { version = "0.0.8", path = "../h3" }
 bytes = "1"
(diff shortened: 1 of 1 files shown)


seanmonstar:
This change isn't needed, since cargo doesn't cap the version to the minimum we set. You can get it by just running `cargo update` in your project.

CosminPerRam:
Bummer, the MSRV of `quinn` has been raised to `1.74.1` (and we are on `1.71.0`), should I make a separate PR for the update? Or this is not wanted here yet?

Issue #164, opened 2023-02-21
CONNECT and tunnel support


stevenctl:
From the proposal:

> HTTP CONNECT to tunnel over a single QUIC stream

Curious what the status is? If nobody is working on this I'd be happy to contribute although I might need a bit of guidance.

nmittler:
Gentle ping ... any feedback from a maintainer would be great.

seanmonstar:
I don't know what the status is. Does it not "just work" right now?

Anything in the spec we likely want to support. If you want to try to figure out what is needed, you're welcome to! I don't know if any of the others have the time to focus on this, though.

nmittler:
@seanmonstar ah got it. I think we were just assuming there were missing pieces, since it was still listed in the roadmap

seanmonstar:
It certainly could be, I haven't checked at all. I've already forgotten what even is required. If it's similar to HTTP/2 CONNECT, then we probably need some work done.

PR #322, opened 2025-11-09
Correct behavior of standard CONNECT

yaroslavros:
Currently h3 client always sets scheme (defaults to `https`) and path (defaults to `/`). This breaks compatibility with CONNECT servers that expect compliance with section 4.4 of RFC9114.

With this change scheme and path are not enforced for standard CONNECT (that is, CONNECT `:method` with no `:protocol` pseudo-header). FWIW, extended CONNECT (that is, CONNECT with some `:protocol`) requires scheme and path.

### DIFF
--- a/h3/src/proto/headers.rs
+++ b/h3/src/proto/headers.rs
@@ -383,6 +383,14 @@ impl Pseudo {
             None
         };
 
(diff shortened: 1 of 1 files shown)


yaroslavros:
This addresses https://github.com/hyperium/h3/issues/164

PR #323, opened 2025-11-12
feat: expose 0-RTT detection at stream level

iadev09:
## Summary
- Adds `is_0rtt()` method to `RequestStream` to detect streams opened during 0-RTT
- Implements `Is0rtt` trait for QUIC stream types
- Exposes 0-RTT detection through the h3-quinn adapter layer

## Motivation
Applications need to detect 0-RTT streams to properly handle replay-attack risks (RFC 8470 § 5.2).  
For example, non-idempotent operations such as **POST**, **PUT**, or **DELETE** should be rejected or handled with extra care on 0-RTT streams.

### DIFF
--- a/h3-quinn/src/lib.rs
+++ b/h3-quinn/src/lib.rs
@@ -374,14 +374,6 @@ impl RecvStream {
             is_0rtt,
         }
     }
-
-    /// Check if this stream has been opened during 0-RTT.
-    ///
-    /// In which case any non-idempotent request should be considered dangerous at the application
-    /// level. Because read data is subject to replay attacks.
-    pub fn is_0rtt(&self) -> bool {
-        self.is_0rtt
-    }
 }
 
 impl quic::RecvStream for RecvStream {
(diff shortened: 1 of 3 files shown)


h3-quinn/src/lib.rs
```diff
@@ -362,6 +383,17 @@ impl RecvStream {
...
+
+	/// Check if this stream has been opened during 0-RTT.
+	///
+	/// In which case any non-idempotent request should be considered dangerous at the application
+	/// level. Because read data is subject to replay attacks.
+    pub fn is_0rtt(&self) -> bool {
+        self.stream
+            .as_ref()
+            .map(|s| s.is_0rtt())
+            .unwrap_or(false)
```
Ruben2424: when `self.stream` is `None` this will always return `false`.
IIRC this can only happen if the user uses the `poll` API of the streams wrong. If so this function would return `false` even for `0rtt` request. 

We could just change this to `expect`.
The better way IMO would be to call `is_0rtt` in the `new` function and store the result in a new field for this function to return.

Ruben2424:
> What if you could peek at specific `&S` type in a `RequestStream`, and thus if you knew it was quinn, you could use the quinn method to look at it.

Could be useful generally. But in this case having a trait could allow libraries also generic over quic transport to use this information for example to automatically return 425 status code. Maybe something for hyper?

iadev09:
For reference, this topic was also discussed in [quinn-rs/quinn#2440](https://github.com/quinn-rs/quinn/pull/2440)
.
In that context we concluded that 0-RTT state is best observed at the stream level.

Ruben2424:
merged it, thanks.

Issue #330, opened 2026-01-21
panic in RecvStream::stop_sending when body is dropped


Deniskore:
Standalone repro reduced from our Salvo HTTP3 handler. Salvo uses Quinn as its HTTP3 backend, so this shows the same h3‑quinn crash without Salvo. Dropping a request body while recv_data is pending triggers a panic in RecvStream stop_sending. Versions used: h3 0.0.8, h3‑quinn 0.0.10, quinn 0.11.9, rustls 0.23.36

Client
```rust
use anyhow::Result;
use bytes::Bytes;
use futures_util::future;
use h3_quinn::Connection as H3QuinnConnection;
use quinn::{ClientConfig, Endpoint, IdleTimeout, TransportConfig};
use rustls::client::danger::ServerCertVerifier;
use std::sync::Arc;
use std::time::Duration;

#[derive(Debug)]
struct SkipServerVerification;

impl SkipServerVerification {
    fn new() -> Arc<Self> {
        Arc::new(Self)
    }
}

impl ServerCertVerifier for SkipServerVerification {
    fn verify_server_cert(
        &self,
        _end_entity: &rustls::pki_types::CertificateDer<'_>,
        _intermediates: &[rustls::pki_types::CertificateDer<'_>],
        _server_name: &rustls::pki_types::ServerName<'_>,
        _ocsp: &[u8],
        _now: rustls::pki_types::UnixTime,
    ) -> Result<rustls::client::danger::ServerCertVerified, rustls::Error> {
        Ok(rustls::client::danger::ServerCertVerified::assertion())
    }

    fn verify_tls12_signature(
        &self,
        message: &[u8],
        cert: &rustls::pki_types::CertificateDer<'_>,
        dss: &rustls::DigitallySignedStruct,
    ) -> Result<rustls::client::danger::HandshakeSignatureValid, rustls::Error> {
        let provider = rustls::crypto::ring::default_provider();
        rustls::crypto::verify_tls12_signature(
            message,
            cert,
            dss,
            &provider.signature_verification_algorithms,
        )
    }

    fn verify_tls13_signature(
        &self,
        message: &[u8],
        cert: &rustls::pki_types::CertificateDer<'_>,
        dss: &rustls::DigitallySignedStruct,
    ) -> Result<rustls::client::danger::HandshakeSignatureValid, rustls::Error> {
        let provider = rustls::crypto::ring::default_provider();
        rustls::crypto::verify_tls13_signature(
            message,
            cert,
            dss,
[...]

PR #331, opened 2026-01-22
Fix stop_sending panic when recv stream is in-flight

Deniskore:
Related to https://github.com/hyperium/h3/issues/330

Fixes the h3-quinn panic when STOP_SENDING is called during an in-flight read by deferring the stop until the stream is available.

It would be great to see fewer expect/unwrap calls in library code over time there are still many in h3.


vedastro48-star (commented):
Nice work

PR #332, opened 2026-02-04
fix: Handle QUIC NO_ERROR in is_h3_no_error()

yanburman:
Extend is_h3_no_error() to also detect QUIC transport layer NO_ERROR (0x0) alongside HTTP/3 H3_NO_ERROR (0x100). Both indicate graceful connection shutdown per RFC 9000 Section 20.1 and RFC 9114.


Ruben2424:
Duplicate to #336.

Issue #325, opened 2025-11-15
ConnectionError::is_h3_no_error() doesn't detect QUIC-layer NO_ERROR in Undefined variant


iadev09:
## Problem

`ConnectionError::is_h3_no_error()` currently only checks for H3-layer `NO_ERROR`, but misses QUIC-layer `NO_ERROR` closes that are wrapped in the `Undefined` variant. This causes graceful connection closes to be logged/handled as errors.

## Real-World Impact

When using h3-quinn, normal client disconnects (e.g., from curl) are logged as warnings instead of debug messages:

```rust
⚠️ H3 connection closed with error: Remote(Undefined(ConnectionClosed(
    ConnectionClose { error_code: NO_ERROR, frame_type: None, reason: b"" }
))) ip:127.0.0.1
```

Notice `error_code: NO_ERROR` - this is a **graceful close**, not an error.

## Current Implementation

`ConnectionError::is_h3_no_error()` in `h3/src/error/error.rs`:

```rust
pub fn is_h3_no_error(&self) -> bool {
    match self {
        ConnectionError::Local { error: LocalError::Application { code: Code::H3_NO_ERROR, .. } } => true,
        ConnectionError::Remote(ConnectionErrorIncoming::ApplicationClose { error_code })
            if *error_code == Code::H3_NO_ERROR.value() => true,
        _ => false,  // ← Undefined variant falls through!
    }
}
```

This only checks `ApplicationClose`, not `Undefined` which contains QUIC-layer errors.

## Related Context

PR #315 fixed the same issue for `StreamError::is_h3_no_error()` by checking the `RemoteTerminate` variant. This is the connection-level equivalent.

## The Challenge

The `Undefined` variant contains a trait object:

```rust
ConnectionErrorIncoming::Undefined(Arc<dyn std::error::Error + Send + Sync>)
```

The error code is buried inside the boxed error (e.g., `quinn::ConnectionError::ConnectionClosed(ConnectionClose { error_code: ... })`), but:
- The concrete type is erased (trait object)
- Fields are not accessible through the `Error` trait
- h3 is transport-agnostic (can't depend on quinn)

## Possible Approaches

I can see a few options, but I'm not sure which aligns with h3's design philosophy:

1. **String parsing** (pragmatic but fragile):
   ```rust
   ConnectionError::Remote(ConnectionErrorIncoming::Undefined(err)) => {
       format!("{:?}", err).contains("NO_ERROR")
   }
   ```

[...]

PR #336, opened 2026-03-17
Model QUIC ConnectionClosed separately from ApplicationClose

iadev09:
## Summary

QUIC transport-level connection closures are currently mapped to `ConnectionErrorIncoming::Undefined`, which erases type information and makes downstream handling rely on debug strings or heuristics.

This change adds a dedicated `ConnectionClosed { error_code }` variant to `ConnectionErrorIncoming` and maps QUIC `ConnectionClosed` errors explicitly instead of treating them as undefined.

QUIC ConnectionClosed(NO_ERROR) represents graceful shutdown and should not be treated as an undefined error.

## What changed

- add `ConnectionErrorIncoming::ConnectionClosed { error_code }`
- map `quinn::ConnectionError::ConnectionClosed(...)` explicitly in `h3-quinn`
- keep `ApplicationClose` strictly HTTP/3-level
- keep `Undefined` as the fallback for errors that still do not have a structured mapping
- treat QUIC `NO_ERROR` as graceful shutdown in `ConnectionError::is_h3_no_error()`
- pass the new variant through the datagram adapter as well

## Why

`ApplicationClose` and QUIC `ConnectionClosed` carry different semantics and belong to different protocol layers.

`ApplicationClose` is an HTTP/3 application close.
`ConnectionClosed` is a QUIC transport close.

This PR keeps those separate instead of collapsing transport shutdown into `Undefined`.

This is also the connection-level equivalent of the fix in #315 for `StreamError::is_h3_no_error()`, which already handles the stream-level case via `RemoteTerminate`.

## Compatibility

This is a public enum API change: `ConnectionErrorIncoming` gains a new `ConnectionClosed` variant.

That is not strictly backwards-compatible for exhaustive matches, but it avoids erasing structured QUIC transport shutdowns behind `Undefined`, and matches the maintainer guidance that this crate can tolerate such updates in practice.

Unlike an open-ended `Other(...)`-style enum, this does not silently remap behavior at runtime; downstream exhaustive matches get a compile-time signal to handle the new transport-level case explicitly.

## Verification

- `cargo test -p h3-quinn connection_closed_is_mapped_to_structured_variant`
- `cargo check -p h3 -p h3-quinn -p h3-datagram`

Refs #325.

### DIFF
--- a/h3-quinn/src/datagram.rs
+++ b/h3-quinn/src/datagram.rs
@@ -92,6 +92,9 @@ fn convert_h3_error_to_datagram_error(
         ConnectionErrorIncoming::ApplicationClose { error_code } => {
             h3_datagram::ConnectionErrorIncoming::ApplicationClose { error_code }
         }
+        ConnectionErrorIncoming::ConnectionClosed { error_code } => {
+            h3_datagram::ConnectionErrorIncoming::ConnectionClosed { error_code }
+        }
         ConnectionErrorIncoming::Timeout => h3_datagram::ConnectionErrorIncoming::Timeout,
         ConnectionErrorIncoming::InternalError(err) => {
             h3_datagram::ConnectionErrorIncoming::InternalError(err)
--- a/h3-quinn/src/lib.rs
+++ b/h3-quinn/src/lib.rs
@@ -112,16 +112,41 @@ fn convert_connection_error(e: quinn::ConnectionError) -> h3::quic::ConnectionEr
                 error_code: application_close.error_code.into(),
             }
         }
+        quinn::ConnectionError::ConnectionClosed(connection_close) => {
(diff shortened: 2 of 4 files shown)


Ruben2424 (approved):
lgtm, thanks

Streetblock (approved):
Reviewed the full patch and exercised it in a downstream merge-test branch. Keeping QUIC transport-level `ConnectionClosed` separate from HTTP/3 `ApplicationClose` preserves the protocol-layer semantics and gives downstream users a reliable way to recognize QUIC `NO_ERROR` without debug-string heuristics.

The focused mapping test and the full `h3` / `h3-quinn` test suites pass locally. I also found no dependency, build-script, workflow, or unsafe-code changes in this PR. Looks good from a downstream integrator's perspective.

PR #339, opened 2026-04-27
frame: drain buffered bytes before propagating QUIC connection error

### LINKED ISSUE #338: `FrameStream::poll_next` / `poll_data` discards already-buffered bytes when QUIC connection error arrives in same recv batch
# `FrameStream::poll_next` / `poll_data` discards already-buffered bytes when QUIC connection error arrives in same recv batch

## Summary

When the underlying QUIC stack delivers stream data **and** a `CONNECTION_CLOSE` (or other connection-level error) in the same recv batch, [`FrameStream::try_recv`](https://github.com/hyperium/h3/blob/master/h3/src/frame.rs) propagates the connection error via `?` before [`FrameDecoder::decode`](https://github.com/hyperium/h3/blob/master/h3/src/frame.rs) gets a chance to consume bytes already sitting in `BufRecvStream::buf`. The application sees `StreamError::ConnectionError(...)` instead of the response it would otherwise have parsed.

The same shape exists in `FrameStream::poll_data`, where partially buffered DATA frame body bytes are stranded.

## Reproducer (sketch)

```text
   QUIC recv batch (single io_uring submission):
     ┌─────────────────────────────────────────┐
     │ STREAM(stream_id=0, FIN=1, data=[H3...]) │
     │ CONNECTION_CLOSE(error_code=H3_NO_ERROR) │
     └─────────────────────────────────────────┘

   Application timeline:
     poll #N   → BufRecvStream::poll_read pulls STREAM bytes into buf
     poll #N+1 → BufRecvStream::poll_read returns Err(ConnectionLost(NO_ERROR))
                 try_recv `?` propagates error
                 decoder.decode never runs against buffered HEADERS frame
                 recv_response → StreamError::ConnectionError
```

This is platform-and-kernel sensitive but reproducible on Linux + io_uring with quinn-udp's batched recv: a fast backend that completes its response and immediately tears down with `H3_NO_ERROR` produces the coalesced datagram that triggers it.

## Why it's a bug

The QUIC layer (quinn-proto / quinn) DOES preserve buffered stream data after CONNECTION_CLOSE — the per-stream assembler is not flushed in `close_common`, and `Chunks::next()` returns buffered chunks before checking connection state. The race is purely at h3's frame layer: `try_recv` assumes `poll_read` is the only source of "more decodable bytes," but on coalesced batches the bytes pulled into `BufRecvStream::buf` on a previous poll are still valid input for the decoder.

`H3_NO_ERROR` is RFC 9114 §8.1's "graceful shutdown without an error" code. A backend that uses it correctly today produces a 502 at the client because of this discarding; the result is functionally equivalent to misclassifying TCP FIN as RST.

## Existing partial fix

[Hyperium PR #339 ](https://github.com/hyperium/h3/pull/339) (commit `dbc9e09` in our downstream tree) recovered the recv_data case for response bodies via a `drain_h3_response_body` helper that re-buffers bytes at the application level and detects "graceful close after complete body." That change works around the symptom one layer up but doesn't fix the underlying frame-layer discard. The recv_response path (no headers parsed yet) cannot be similarly worked around because the application has no buffered bytes at that point — only h3 does.

## Proposed fix

Hoist the QUIC error out of `try_recv`'s `?` path. When `poll_read` returns an error, cache it for the current iteration; let `decoder.decode(self.stream.buf_mut())` run once. If the decoder produces a frame, return it normally. If the decoder returns `None` from buffered bytes, surface the cached error.

Same shape in `poll_data`: don't strand partial body bytes.

The fix terminates: each iteration consumes its own cached error, so partial-frame buffers cannot trigger an infinite loop. Forward progress is bounded by the size of the buffered bytes, which can only shrink.

A working diff plus two new regression tests (using an extended `FakeRecv` that supports `chunk_then_error`) accompanies this issue as PR. The tests exercise both the headers and body paths against a synthetic `ConnectionErrorIncoming::ApplicationClose { error_code: 0x100 }`.

## Impact

- HTTP/3 gateways / clients on Linux + io_uring see spurious 502s for backends that use `H3_NO_ERROR` correctly (per-connection request limits, drain/rolling deploys, stateless backends).
- Protocols/products affected: any consumer of `RequestStream::recv_response()` and `recv_data()` on the client side; the analogous server-side path likely has the same shape.
- Severity: clean shutdowns mis-classified as transport errors — wrong behavior, but worked around in some downstream stacks (e.g., ours via [our recv_response gateway-side suppression](https://github.com/ferrum-edge/ferrum-edge/pull/506)) by surfacing the 502 without penalizing the backend's H3 capability.

## Versions

- `h3 = 0.0.8`
- `h3-quinn = 0.0.10`
- `quinn = 0.11.9`
- `quinn-proto = 0.11.14`

But the bug is structural — same shape exists on `master`. The proposed PR forward-ports cleanly.

## References

- RFC 9114 §8.1 — H3 error codes (`H3_NO_ERROR = 0x100`)
- RFC 9000 §10.2 — QUIC connection close
- Downstream context (Ferrum Edge): [PR #506](https://github.com/ferrum-edge/ferrum-edge/pull/506) and [PR #505](https://github.com/ferrum-edge/ferrum-edge/pull/505) handle the symptom; this issue is the root-cause report.

jeremyjpj0916:
# frame: drain buffered bytes before propagating QUIC connection error

Fixes #338.

## Problem

`FrameStream::try_recv` propagates a connection-level error from
`BufRecvStream::poll_read` via `?` before the frame decoder gets to run
against bytes that may already be sitting in `BufRecvStream::buf` from
a previous wake. When a QUIC stack delivers stream data and a
`CONNECTION_CLOSE` (or other connection error) in the same recv batch —
common on Linux with io_uring's batched recvmsg and coalesced UDP
datagrams — buffered HEADERS / DATA bytes are discarded and clients
see `StreamError::ConnectionError` instead of the response that was
already on the wire.

The QUIC layer is innocent here: quinn-proto's per-stream assembler is
not flushed by `close_common`, and `Chunks::next()` returns buffered
chunks before checking connection state. The discard is purely at
h3's frame layer.

`H3_NO_ERROR` is RFC 9114 §8.1's "graceful shutdown without an error"
code. Today, a backend that uses it correctly (per-connection request
limits, drain/rolling deploy, stateless backends) produces 502s at the
client.

## Fix

Hoist the QUIC error out of `try_recv`'s `?` path. When `poll_read`
returns an error, cache it for the current iteration; let
`decoder.decode(self.stream.buf_mut())` run once against whatever is
already buffered. If the decoder produces a frame, surface it normally.
If `decoder.decode` returns `None`, surface the cached error.

Same shape applied to `FrameStream::poll_data` so DATA frame body
bytes that were buffered before the error are not stranded.

### Termination

Each iteration consumes its own cached error in the same iteration it
was observed. A partial-frame buffer that the decoder cannot resolve
exits the loop with the cached error in the `None` arm of the match —
no opportunity for an infinite loop.

### Forward progress

When the decoder DOES produce a frame on the cached-error iteration,
the next poll re-issues `poll_read`, which returns the same error.
The buffered bytes have shrunk by the consumed frame's length, so
iterations are bounded by the buffer size.

## Tests

Two new tests in `frame.rs::tests`:

- `poll_next_drains_buffered_headers_before_quic_close` — synthesizes a
  HEADERS frame followed by an `ApplicationClose { error_code: 0x100 }`
  via `FakeRecv::chunk_then_error`. First poll yields the frame, second
  poll surfaces the error.
- `poll_data_drains_buffered_body_before_quic_close` — same shape on
  the body path. `poll_next` yields the DATA frame header, `poll_data`
  drains the buffered 4-byte body even though the next poll would
  return the connection error.

`FakeRecv` is extended with a `chunk_then_error` helper that queues a
synthetic `StreamErrorIncoming::ConnectionErrorIncoming` for the next
poll after a chunk drains. This is the smallest model of the recv-batch
race that a unit test can construct without a live QUIC backend.

The existing tests in `frame.rs` still pass (the change is a no-op for
the non-error path).

## Compatibility

- Public API: unchanged. `poll_next` and `poll_data` keep the same
  signatures and the same return-shape contract — successful frames
  continue to win, errors continue to surface, the only difference is
  the ordering when both are pending.
- Internal API: no callers outside `frame.rs` are affected.
- Behavioral compat: a backend that intentionally aborts a stream with
  a non-`NO_ERROR` code mid-frame will now have its buffered partial
  data delivered before the error fires. This is arguably more correct
  (the bytes WERE received before the abort) but worth flagging in case
  any downstream depends on the prior "discard on error" semantic.

## Downstream context

This is the root-cause fix for the recv_response analog of [#NNN
(predecessor PR for recv_data graceful close)]. The recv_data path was
worked around at the application layer in commit `dbc9e09` via a
`drain_h3_response_body` helper that re-implemented buffer-aware
recovery one level up. With this PR, that helper can be simplified —
the underlying decoder will already drain on its own — but I've kept
the helper unmodified to keep the change here minimal. Happy to follow
up with simplification once this lands if maintainers prefer.

The recv_response path is the one this PR specifically fixes: at that
boundary the application has no buffered bytes of its own to fall back
on, so the only fix is at h3's frame layer.

## Refs

- RFC 9114 §8.1 — H3 error codes
- Issue #338 — bug report with reproducer and rationale
- Downstream consumer: [ferrum-edge#506](https://github.com/ferrum-edge/ferrum-edge/pull/506) — gateway-side suppression of capability-downgrade for graceful closes (treats the symptom while we await this upstream fix)

### DIFF
--- a/h3/src/frame.rs
+++ b/h3/src/frame.rs
@@ -70,7 +70,17 @@ where
         );
 
         loop {
-            let end = self.try_recv(cx)?;
+            // Hoist the QUIC error out of `?` so that any bytes pulled
+            // into `self.stream.buf_mut()` on prior wakes get one last
+            // chance through the decoder. Without this hoist, a
+            // CONNECTION_CLOSE frame coalesced with stream data in the
+            // same recv batch wins the race and discards a fully-
+            // decodable HEADERS frame.
+            let (end, pending_quic_err) = match self.try_recv(cx) {
+                Poll::Ready(Ok(end)) => (Poll::Ready(end), None),
+                Poll::Pending => (Poll::Pending, None),
+                Poll::Ready(Err(e)) => (Poll::Ready(true), Some(e)),
+            };
 
             return match self.decoder.decode(self.stream.buf_mut())? {
                 Some(Frame::Data(PayloadLen(len))) => {
@@ -82,11 +92,17 @@ where
                     Poll::Ready(Ok(frame))
                 }
                 Some(frame) => Poll::Ready(Ok(Some(frame))),
-                None => match end {
+                None => match (end, pending_quic_err) {
+                    // Decoder couldn't make progress on buffered bytes
+                    // and the underlying QUIC stream errored: surface
+                    // the cached error now. Fixed-point: this branch
+                    // terminates the loop because we consume the cached
+                    // error in the same iteration we observed it.
+                    (_, Some(err)) => Poll::Ready(Err(err)),
                     // Received a chunk but frame is incomplete, poll until we get `Pending`.
-                    Poll::Ready(false) => continue,
-                    Poll::Pending => Poll::Pending,
-                    Poll::Ready(true) => {
+                    (Poll::Ready(false), None) => continue,
+                    (Poll::Pending, None) => Poll::Pending,
+                    (Poll::Ready(true), None) => {
                         if self.stream.buf_mut().has_remaining() {
                             // Reached the end of receive stream, but there is still some data:
                             // The frame is incomplete.
@@ -112,23 +128,32 @@ where
             return Poll::Ready(Ok(None));
         };
 
-        let end = match self.try_recv(cx) {
-            Poll::Ready(Ok(end)) => end,
-            Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
-            Poll::Pending => false,
+        // Mirror the hoist from poll_next: a QUIC-level error must not
+        // discard already-buffered body bytes for the current frame.
+        let (end, pending_quic_err) = match self.try_recv(cx) {
+            Poll::Ready(Ok(end)) => (end, None),
+            Poll::Ready(Err(e)) => (true, Some(e)),
+            Poll::Pending => (false, None),
         };
         let data = self.stream.buf_mut().take_chunk(self.remaining_data);
 
-        match (data, end) {
-            (None, true) => Poll::Ready(Ok(None)),
-            (None, false) => Poll::Pending,
-            (Some(d), true)
-                if d.remaining() < self.remaining_data
+        match (data, end, pending_quic_err) {
+            // No more buffered body and QUIC errored: surface the error.
+            (None, _, Some(err)) => Poll::Ready(Err(err)),
+            (None, true, None) => Poll::Ready(Ok(None)),
+            (None, false, None) => Poll::Pending,
+            // Partial body for the current frame and the underlying
+            // stream / connection ended: caller MUST treat this as
+            // truncation. Same shape as the existing `Poll::Ready(true)`
+            // branch below — extended to also cover the QUIC error case.
+            (Some(d), end_or_err, pending)
+                if (end_or_err || pending.is_some())
+                    && d.remaining() < self.remaining_data
                     && !self.stream.buf_mut().has_remaining() =>
             {
                 Poll::Ready(Err(FrameStreamError::UnexpectedEnd))
             }
-            (Some(d), _) => {
+            (Some(d), _, _) => {
                 self.remaining_data -= d.remaining();
                 Poll::Ready(Ok(Some(d)))
             }
@@ -323,6 +348,7 @@ mod tests {
     use std::collections::VecDeque;
 
     use crate::proto::{coding::Encode, frame::FrameType, varint::VarInt};
+    use crate::quic::ConnectionErrorIncoming;
 
     // Decoder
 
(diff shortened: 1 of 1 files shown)


Streetblock (commented):
Thanks for working on this. While testing the patch downstream, I found one error-propagation case that the current tests do not cover.

In `poll_data`, when `try_recv` returns an error while a complete DATA body is already buffered, the body is returned from the `(Some(d), _, _)` branch, but `pending_quic_err` is local to that poll and is then dropped. With the PR's `FakeRecv`, the error was already consumed through `pending_error.take()`, so the next `poll_next` returns `Ok(None)` instead of the connection error.

Adding this assertion after the existing body assertion reproduces it:

```rust
assert_poll_matches!(
    |cx| stream.poll_next(cx),
    Err(FrameStreamError::Quic(_))
);
```

It fails with `Ok(None)`.

The HEADERS regression test also does not quite exercise the stated “buffered bytes and error observed in the same poll” case: its first poll reads and decodes the queued chunk, while its second poll observes the error with an empty buffer.

I tested a fix that stores the deferred error in `FrameStream`, drains buffered frames/body data, and then surfaces the saved error exactly once. The focused regressions and the complete `h3` / `h3-quinn` suites pass. The downstream commit is: https://github.com/Streetblock/h3/commit/b24e0af

I think the deferred error should be preserved across polls before this is merged.

Issue #335, opened 2026-03-08
Direct path import of h3-quinn in test


alexanderkjall:
Here h3-quinn is pulled in by a path reference: https://github.com/hyperium/h3/blob/master/h3/src/tests/mod.rs#L8

This makes it hard to run the test-suite in Debian, as we don't layout the files in the exact same way as you do in this git repository.

Would it be possible to have this as a normal `[dev-dependencies]` instead?

seanmonstar:
The comments right above it explain why it isn't currently a dev-dependency. 🤷‍♂️

PR #340, opened 2026-05-18
fix(datagram): preserve quarter stream id

HeikoBornholdt:
`h3-datagram` encoded the Quarter Stream ID correctly, but then stored a zeroed buffer in `EncodedDatagram`. As a result, every outgoing HTTP/3 Datagram was serialized with Quarter Stream ID `0`.

This breaks datagrams for later request streams on the same HTTP/3 connection, for example when multiple MASQUE CONNECT-UDP sessions share one H3 connection. Those datagrams appear to belong to the first request stream and may be routed to the wrong session, ignored, or rejected.

### DIFF
--- a/h3-datagram/src/datagram.rs
+++ b/h3-datagram/src/datagram.rs
@@ -72,7 +72,7 @@ where
         let varint = VarInt::from(self.stream_id) / 4;
         varint.encode(&mut buffer.as_mut_slice());
         EncodedDatagram {
-            stream_id: [0; VarInt::MAX_SIZE],
+            stream_id: buffer,
             len: varint.size(),
             pos: 0,
             payload: self.payload,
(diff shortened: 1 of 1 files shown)


Streetblock (approved):
Tested this commit in an isolated merge against current `upstream/master`.

- the merge is conflict-free
- `cargo fmt --all -- --check` passes
- `cargo test --workspace` passes (227 h3 tests, the h3-datagram test, and 9 documentation tests)
- the only warning is the pre-existing direct-path `h3-quinn` cfg warning tracked separately in #335

I also expanded the round-trip test locally across all QUIC VarInt width boundaries for Quarter Stream IDs (`0`, `1`, `2^6-1`, `2^6`, `2^14-1`, `2^14`, `2^30-1`, `2^30`, and `2^60-1`). All cases preserve the original stream ID and payload.

The one-line production fix correctly uses the already encoded header buffer instead of replacing it with zeros. The broader boundary table could be a useful optional test enhancement, but I do not consider it a merge blocker.

seanmonstar (approved):
Looks good, thanks!

PR #341, opened 2026-06-11
fix: h3 datagrams must not be sent until both sides agree on it

JPDye:
> https://www.rfc-editor.org/rfc/rfc9297#section-2.1.1:
QUIC DATAGRAM frames MUST NOT be sent until the SETTINGS_H3_DATAGRAM setting has been both sent and received with a value of 1.

Changed the `SharedState` so that `enable_datagram` represents the negotiated SETTINGS_H3_DATAGRAM (an AND of ours and the remotes).

Also fixed some broken duvet comments and added duvet tracking of h3-datagram and the associated RFC.

### DIFF
--- a/.duvet/config.toml
+++ b/.duvet/config.toml
@@ -6,6 +6,12 @@ pattern = "h3/**/*.rs"
 [[specification]]
 source = "https://www.rfc-editor.org/rfc/rfc9114"
 
+[[source]]
+pattern = "h3-datagram/**/*.rs"
+
+[[specification]]
+source = "https://www.rfc-editor.org/rfc/rfc9297"
+
 [report.html]
 enabled = true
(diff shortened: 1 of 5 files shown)


vcarus:
Two of the RFC 9297 §2.1.1 lines this PR adds to the duvet snapshot ("The value of the SETTINGS_H3_DATAGRAM setting MUST be either 0 or 1" and "... MUST terminate the connection with error H3_SETTINGS_ERROR") are cheap to implement while you are here: `Settings::decode` in `h3/src/proto/frame.rs` never looks at a setting's value, and `SettingsError::InvalidSettingValue(SettingId, u64)` is already defined with a `Display` impl but is never constructed anywhere in the crate. A check for `SettingId::H3_DATAGRAM` with a value other than 0 or 1 returning that variant would let both citations be marked as implemented instead of registered.

PR #342, opened 2026-06-15
fix: encode HTTP/3 datagram QSID

JPDye:
The datagram Quarter Stream ID is constructed but never actually used when encoding. Now it is.


JPDye:
Closing. I realise #340 solves this.

PR #343, opened 2026-07-06
Update documentation of `recv_data()`/`poll_recv_data()`

divergentdave:
This updates the documentation of various `recv_data()` and `poll_recv_data()` methods. My primary confusion was that they only return a frame of the body at a time, so I made mention of that in the methods that didn't make it explicit yet. There were also some incorrect references to requests and responses, so I cleaned those up.

### DIFF
--- a/h3/src/client/stream.rs
+++ b/h3/src/client/stream.rs
@@ -184,14 +184,18 @@ where
         Ok(resp)
     }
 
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):
Nice improvement, thanks!

PR #344, opened 2026-08-05
Apply backpressure before reading more frames

### LINKED ISSUE #308: poll_read data is buffered without bounds
https://github.com/hyperium/h3/blob/84316da426d1825a871cd850ff4f90aa3fe5860b/h3/src/stream.rs#L433

The `poll_read()` on `BufRecvStream` (wrapping the underlying QUIC provider's stream) is buffering chunks in-memory in a `BufList` that appears to grows without bounds:

https://github.com/hyperium/h3/blob/master/h3/src/stream.rs#L448-L450

In some tests that I ran over loopback to send and receive data as fast as possible, if the client does not consume from `BufRecvStream` quickly (for any reason) I observed that the client memory grows without bounds.  This should ideally not be the case -- a slow consumer should back-pressure the sender.  This problem does not exist if I use the underlying QUIC provider's stream directly.

I don't have a handy reproducer that I can share at the moment, but I can work on one if it'll be useful.

Streetblock:
## Summary

Decode complete frames already present in `FrameStream`'s receive buffer before polling the underlying QUIC transport for another chunk.

This prevents a slow frame consumer from causing `BufRecvStream` to accumulate a new transport chunk on every wake while complete frames are still buffered.

## Root cause

`FrameStream::poll_next` called `try_recv` before attempting to decode the existing buffer. If one transport chunk contained multiple HTTP/3 frames, every subsequent frame poll could read and append another chunk before consuming the frames already available.

## Regression test

The new deterministic test models 64 immediately available transport chunks, each containing 16 HEADERS frames with a 1,024-byte payload, while the consumer processes one frame per wake.

On current `master`, the test observed:

- 64 underlying transport polls
- 985,920 bytes left buffered after consuming 64 frames
- one encoded transport chunk is 16,432 bytes

With this change:

- only 4 transport polls are needed for the same 64 frames
- buffering remains bounded to the currently consumed transport chunk

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p h3 -p h3-quinn`
- 228 unit tests passed
- 9 documentation tests passed

## Prior and related work

After opening this draft, I noticed the issue timeline links to [0x676e67/http3#2](https://github.com/0x676e67/http3/pull/2). That June 2026 fork patch already contains the equivalent `FrameStream` reordering and a smaller two-frame regression test, alongside unrelated QPACK changes. Credit to @0x676e67 for that earlier implementation.

This draft was developed independently against current upstream `master`; its additional contribution is the stronger multi-chunk regression scenario and measured baseline above. Maintainers may prefer to adapt the earlier commit or this focused patch, and I am happy to adjust attribution or approach accordingly.

The same regression scenario also passes on the broader receive-buffer refactor in #328. This draft offers a smaller near-term change while that larger design is evaluated.

Fixes #308

### DIFF
--- a/h3/src/frame.rs
+++ b/h3/src/frame.rs
@@ -70,33 +70,33 @@ where
         );
 
         loop {
-            let end = self.try_recv(cx)?;
-
-            return match self.decoder.decode(self.stream.buf_mut())? {
+            match self.decoder.decode(self.stream.buf_mut())? {
                 Some(Frame::Data(PayloadLen(len))) => {
                     self.remaining_data = len;
-                    Poll::Ready(Ok(Some(Frame::Data(PayloadLen(len)))))
+                    return Poll::Ready(Ok(Some(Frame::Data(PayloadLen(len)))));
                 }
                 frame @ Some(Frame::WebTransportStream(_)) => {
                     self.remaining_data = usize::MAX;
-                    Poll::Ready(Ok(frame))
+                    return Poll::Ready(Ok(frame));
                 }
-                Some(frame) => Poll::Ready(Ok(Some(frame))),
-                None => match end {
-                    // Received a chunk but frame is incomplete, poll until we get `Pending`.
-                    Poll::Ready(false) => continue,
-                    Poll::Pending => Poll::Pending,
-                    Poll::Ready(true) => {
-                        if self.stream.buf_mut().has_remaining() {
-                            // Reached the end of receive stream, but there is still some data:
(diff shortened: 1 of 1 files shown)


seanmonstar:
This looks good to me, reason it's still a draft?

Streetblock:
Oops, my bad! Leaving it as a draft was just an oversight. I just marked it as ready for review. Thanks!

seanmonstar (approved):

PR #362, opened 2026-08-23
Put in a smidge of elbow grease to reduce excess GREASE

arrimorris:
Currently, the `send_grease_frame` flag is only cleared inside `Connection::accept()`. However, third-party backends and embedders (like `hyper`) that require fine-grained async control use the i-implement-a-third-party-backend... feature to bypass `accept()` and drive the connection via `poll_accept_request_stream_internal()`. Since embedders bypass accept(), the send_grease_frame flag is never cleared. No bueno.

By moving this toggle down into `create_resolver_internal`, the GREASE frame is correctly sent once per connection, regardless of whether we call the high-level accept() or the low-level poll_accept_request_stream_internal() API.


Ruben2424 (approved):
lgtm

